PR #19, opened 2020-10-14
AES-GCM

andreistefanescu:
Add AES-GCM proof (with AES-256).


SAW/proof/AES/AESNI-GCM.saw
```diff
@@ -0,0 +1,86 @@
...
+  crucible_points_to_untyped (crucible_elem Xi_ptr 0) (crucible_term {{ (cipher_update enc ctx (take`{res_len} in_)).Xi }} );
+
+  crucible_return (crucible_term {{ `res_len : [64] }});
+};
+
+
+////////////////////////////////////////////////////////////////////////////////
+// Proof commands
+
+let aesni_gcm_cipher_tactic = do {
```
apetcher-amazon, line 49: If possible, but tactics and other proof elements in a separate file. Add comments to state that these elements do not need to be inspected.

SAW/proof/AES/goal-rewrites.saw
```diff
@@ -0,0 +1,340 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+*/
+
+
```
apetcher-amazon, line 6: Add a comment stating that this file is all proof (no spec/assumption), so it doesn't require inspection in order to trust the result.

SAW/proof/AES/AES-GCM.saw
```diff
@@ -0,0 +1,255 @@
...
+
+include "goal-rewrites.saw";
+
+
+m <- llvm_load_module "../../build/llvm/crypto/crypto_test.bc";
+
+
+let NID_aes_256_gcm = 901;
+let aes_block_size = 1;
+let aes_iv_len = 12;
```
apetcher-amazon, line 24: Add a caveat to the README files that states that the proof is only valid 12-byte IVs.

apetcher-amazon (commented):
See specific comments in files.

SAW/proof/AES/AES.saw
```diff
@@ -0,0 +1,122 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+*/
+
+
+////////////////////////////////////////////////////////////////////////////////
+// Specifications
+
+let AES_BLOCK_SIZE = 16;
+let aes_key_len = 32;
+let aes_rounds = 13;
```
apetcher-amazon, line 12: This name is a bit confusing (it's probably confusing in the BoringSSL code, too). In general, should this be number of rounds - 1? Describe the situation with a clearer variable name or a comment.

apetcher-amazon (commented):
First round of fixes look great---especially the Cryptol refactoring. Just one minor comment about confusing names in BoringSSL code.

SAW/spec/AES/AES-GCM.cry
```diff
@@ -0,0 +1,93 @@
...
+
+/*
+ * AES specifications.
+ */
+aes_key_to_schedule : [Nr + 1][128] -> KeySchedule
+aes_key_to_schedule key = (key' @ 0, key' @@ [1 .. (Nr - 1)], key' @ `Nr)
+  where
+    key' = map msgToState_x86 key
+
+aes_key_from_schedule : KeySchedule -> [4 * (Nr - 1)][32]
```
apetcher-amazon, line 34: Should this be in AES spec file(not specific to GCM)?

SAW/proof/AES/AES-GCM.saw
andreistefanescu: I think AES-GCM requires the size of the counter to be 4 bytes, and thus the size of the IV is (AES block size) - (counter size) = 16 - 4 = 12.

SAW/proof/AES/AES-GCM.saw
```diff
@@ -0,0 +1,255 @@
...
+  crucible_points_to (crucible_field ptr "ivlen") (crucible_term {{ `aes_iv_len : [32] }});
+  crucible_points_to (crucible_field ptr "taglen") (crucible_term {{ `taglen : [32] }});
+  crucible_points_to (crucible_field ptr "iv_gen") (crucible_term {{ 0 : [32] }});
+  crucible_points_to (crucible_field ptr "ctr") (crucible_global "aes_hw_ctr32_encrypt_blocks");
+};
+
+
+let aes_gcm_from_cipher_ctx_spec = do {
+  ctx_ptr <- crucible_alloc_readonly (llvm_struct "struct.evp_cipher_ctx_st");
+  cipher_data_ptr <- crucible_fresh_pointer (llvm_struct "struct.EVP_AES_GCM_CTX");
```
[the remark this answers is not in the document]
andreistefanescu: good catch!

SAW/spec/AES/AES-GCM.cry
andreistefanescu: this is used only in `points_to_aes_key_st`, in order to specify the part of the key schedule that is derived from the AES key. The order of the bytes is specific to this implementation.

SAW/proof/AES/AES-GCM.saw
apetcher-amazon: Right, but there is an extra step at the beginning that allows IVs of other sizes to be used. If the IV is not 12 bytes long, it is hashed in a particular way to produce a 12-byte IV. See step 2 of page 15 of [SP 800-38d](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38d.pdf).

Another option is to add another case that verifies on some IV length other than 12. But this is probably not worth it, since using these other IV lengths in applications is generally discouraged.

SAW/spec/AES/AES-GCM.cry
apetcher-amazon: I see. I missed the implementation-specific part. Still seems like it should be in a file containing AES definitions (maybe AES/X86.cry), but it's fine to leave it where it is, for now. I expect to refactor a bit so I can reuse definitions like this in the AES KW proof.

SAW/proof/AES/AESNI-GCM.saw
```diff
@@ -0,0 +1,86 @@
...
+  simplify (addsimps xor_slice_append_thms basic_ss);
+  simplify (addsimps concat_assoc_0_thms empty_ss);
+  simplify (addsimps concat_assoc_1_thms empty_ss);
+  simplify (addsimps concat_assoc_2_thms empty_ss);
+  simplify (addsimps [cmp_sub_thm] empty_ss);
+  w4_unint_yices ["pmult", "pmod", "gcm_polyval", "aesEncryptWithKeySchedule"];
+};
+
+let do_prove = false;
+add_x86_preserved_reg "rax";
```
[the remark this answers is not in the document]
apetcher-amazon: I need a bit more information in the comment to understand what add_x86_preserved_reg does. Add a sentence or two describing what this statement does and why it is needed in this case. Most importantly: does this statement introduce new assumptions?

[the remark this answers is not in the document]
andreistefanescu: x86 verification has a "code discovery" pre-processing phase,  when x86 instructions are organized in basic blocks and functions. Then verification is performed on functions and basic blocks. Code discovery uses abstract interpretation to track jump addresses and return addresses. `add_x86_preserved_reg "rax"` tracks `%rax` in the abstract domain across function calls (`%rax` is caller-saved by ABI compliant functions, and by default the abstract domain throws away any information about caller-saved registers). This does not introduce any new assumptions.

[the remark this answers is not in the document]
apetcher-amazon: Can you add a shorter version of this explanation as a comment in the code? Something like "The following command tells the analysis to track %rax across function calls. This statement does not introduce any new assumptions."

SAW/proof/AES/AES-GCM.saw
andreistefanescu: as far as I can see, for AES-GCM, `EVP_CipherInit_ex`  (https://github.com/google/boringssl/blob/fd83592b425446444525f30537a577ca2e069c20/crypto/fipsmodule/cipher/cipher.c#L134) expects a 12-byte IV. Hashing the IV to 12 bytes would happen before `EVP_CipherInit_ex`, and is not included in this proof.

apetcher-amazon: I looked again at the code, and I think you're right---this function only supports 12-byte IVs. The implementation might support other IV lengths, but that would require the use of a function like EVP_CIPHER_CTX_ctrl. Since that API function isn't verified, it is implied that the verification only covers 12-byte IVs.

andreistefanescu:
@apetcher-amazon I didn't notice that the patch was failing. That what was causing the OOM. Now that the patch is fixed, all proof run in github actions.

apetcher-amazon (approved):

PR #35, opened 2021-01-15
Update SAW version

chameco:

This pull request updates the SAWScript nightly version to include support for the `pinsrd` instruction, which is necessary to verify `aes_hw_ctr32_encrypt_blocks` for `len > 1`.

`aes_hw_ctr32_encrypt_blocks` branches on `len` to specially handle the case where `len = 1`, which we were previously verifying. SAWScript's binary analysis could not identify the `pinsrd` instruction used in the other branch (when `len > 1`), and therefore translated the subsequent instructions in the branch as an "error" primitive. Now that SAWScript has support for `pinsrd`, the rest of the branch is properly translated. However, this has an unfortunate side effect: `aes_hw_ctr32_encrypt_blocks` saves the original stack pointer in `r11`, which is a caller-saved register. The newly-translated branch contains calls to helper routines (e.g. to `.Lenc_loop8_enter`). Although these helper routines do not overwrite `r11`, this situation prevents the analysis from identifying the return at the end of the routine. Locally treating `r11` as callee-saved fixes this.

We've updated the specification for `aes_hw_ctr32_encrypt_blocks` to deal with the above, using a slightly modified version of the multi-block specification from the `aes-saw` branch.

### DIFF
--- a/SAW/proof/AES/AES.saw
+++ b/SAW/proof/AES/AES.saw
@@ -177,10 +177,19 @@ aes_hw_decrypt_in_place_ov <- llvm_verify_x86 m "../../build/x86/crypto/crypto_t
   });
 
 add_x86_preserved_reg "r11";
-aes_hw_ctr32_encrypt_blocks_ov <- llvm_verify_x86 m "../../build/x86/crypto/crypto_test" "aes_hw_ctr32_encrypt_blocks"
+llvm_verify_x86 m "../../build/x86/crypto/crypto_test" "aes_hw_ctr32_encrypt_blocks"
   []
   true
   (aes_hw_ctr32_encrypt_blocks_spec 2)
+  (do {
+    simplify (cryptol_ss ());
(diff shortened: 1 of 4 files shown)


SAW/proof/AES/AES.saw
```diff
@@ -173,6 +176,16 @@ aes_hw_decrypt_in_place_ov <- llvm_verify_x86 m "../../build/x86/crypto/crypto_t
     w4_unint_yices ["AESInvRound", "AESFinalInvRound"];
   });
 
+add_x86_preserved_reg "r11";
```
apetcher-amazon, line 179: Include a comment that describes what the code does, and how the analysis handles it. This was described well in the pull request description---we just need similar documentation in the code.

apetcher-amazon (commented):
Just one small documentation request. Looks good otherwise.

apetcher-amazon (approved):

PR #36, opened 2021-02-03
Add proof for RSA-PSS. 

andreistefanescu:


SAW/proof/common/internal.saw
```diff
@@ -66,6 +66,24 @@ let ERR_put_error_spec = do {
     ];
 };
 
+// TODO: Override the functions that call into these CRYPTO_MUTEX functions,
```
apetcher-amazon, line 69: Please make this change as part of the PR, or remove the TODO and open a ticket after the PR in merged.

SAW/proof/RSA/RSA.saw
```diff
@@ -0,0 +1,577 @@
...
+  crucible_points_to (crucible_field ptr "mont_q") mont_q_ptr;
+  crucible_points_to (crucible_field ptr "dmp1_fixed") crucible_null;
+  crucible_points_to (crucible_field ptr "dmq1_fixed") crucible_null;
+  crucible_points_to (crucible_field ptr "inv_small_mod_large_mont") inv_small_mod_large_mont_ptr;
+
+  crucible_points_to (crucible_field ptr "mont_n") mont_n_ptr;
+  // Must be null for ensure_fixed_copy
+  crucible_points_to (crucible_field ptr "d_fixed") crucible_null;
+
+  // TODO: Setting this to zero skips the blinding cache invalidation step for
```
apetcher-amazon, line 126: Only need to check the default case. I.e. the value of this field when the context is created.

apetcher-amazon (changes requested):
In addition to the detailed comments:

1) It looks like RSA key pair generation is not verified. Can you add a proof for this function? 
2) Please update the README files describing the proofs.
3) Need to add a line to the run_checks script to check the new RSA proof.

andreistefanescu:
@apetcher-amazon the RSA proof requires the latest SAW, I am working on updating SAW in the CI.

apetcher-amazon (changes requested):
After reading through the code and README, I've come to the conclusion that this proof is a good start, but I don't think it is sufficiently "done" to include it in the top-level README files. I worry that someone will misunderstand what is verified, and will expect that the key generation function or RSA-PSS at higher than 1024 bits is verified. From our offline discussions, I get the impression that the best approach is to attempt a proof of RSA 3072 after some tool improvements.

Can you take the text about RSA out of the top-level README files and put it in a README file in the RSA directory? That README should clearly state that this is a work in progress.

apetcher-amazon (approved):

PR #40, opened 2021-02-26
Generalize gcm_ghash_avx proof.

andreistefanescu:


andreistefanescu:
@apetcher-amazon this works for up to 17 blocks, except for 7 and 15, which fail because the SAW x86_64 semantics is incomplete. @chameco is fixing that, and then I can update SAW in CI.

apetcher-amazon:
> @apetcher-amazon this works for up to 17 blocks, except for 7 and 15, which fail because the SAW x86_64 semantics is incomplete. @chameco is fixing that, and then I can update SAW in CI.

Thanks. I would prefer to use this PR to bring in the new version of SAW and get the proof working for all lengths up to 17 blocks. So you can update this PR with the new SAW version, when it is ready.

apetcher-amazon (approved):

PR #47, opened 2021-04-08
Test 105 len in GitHub CI.

bryce-shang:
This PR will be closed when GitHub CI finished the job.


bryce-shang:
Confirmed AES-GCM 105 case.

PR #48, opened 2021-04-28
Add SHA-512 specifications with SMT arrays.

andreistefanescu:

### DIFF
--- a/SAW/proof/SHA512/SHA512-common.saw
+++ b/SAW/proof/SHA512/SHA512-common.saw
@@ -115,6 +115,8 @@ let sha512_block_data_order_array_spec = do {
   let len = {{ num * `SHA512_CBLOCK }};
   (data, data_ptr) <- ptr_to_fresh_array_readonly "data" len;
 
+  crucible_precond {{ len != 0 }};
+
   // Call function with `state_ptr`, `data_ptr`, and `num`
   crucible_execute_func [state_ptr, data_ptr, (crucible_term num)];
 
@@ -206,7 +208,8 @@ let fresh_sha512_state_st_array name = do {
   // Message block
   block <- crucible_fresh_cryptol_var (str_concat name ".block") {| ByteArray |};
   // Index
-  n <- crucible_fresh_var (str_concat name ".n") i32;
+  n' <- crucible_fresh_var (str_concat name ".n") i8;
+  let n = {{ (0 # n') : [32] }};
   // Size
   sz <- crucible_fresh_var (str_concat name ".sz") i128;
   // Build SHAState
@@ -226,13 +229,13 @@ let fresh_sha512_state_st_array name = do {
  */
 // Specify the sha512_state_st_array struct from a SHAState
 let points_to_sha512_state_st_common_array ptr (h, sz, block, n) = do {
-  crucible_points_to (crucible_field ptr "h") (crucible_term h);
+  crucible_points_to_array_prefix (crucible_field ptr "p") block {{ `SHA512_CBLOCK : [64] }};
 
   // Specify `sha512_state_st.Nl` and `sha512_state_st.Nh` contain `sz`
   //llvm_points_to_at_type (crucible_field ptr "Nl") i128 (crucible_term sz);
   llvm_points_to_untyped (crucible_field ptr "Nl") (crucible_term sz);
 
-  crucible_points_to_array_prefix (crucible_field ptr "p") block {{ `SHA512_CBLOCK : [64] }};
+  crucible_points_to (crucible_field ptr "h") (crucible_term h);
 
   crucible_points_to (crucible_field ptr "num") (crucible_term n);
   crucible_points_to (crucible_field ptr "md_len") (crucible_term {{ `SHA_DIGEST_LENGTH : [32] }});
@@ -240,7 +243,7 @@ let points_to_sha512_state_st_common_array ptr (h, sz, block, n) = do {
 
 // Specify the sha512_state_st struct from a SHAState
 let points_to_sha512_state_st_array ptr state = do {
- points_to_sha512_state_st_common_array ptr ({{ state.h }}, {{ state.sz }}, {{ state.block }}, {{ state.n }});
+  points_to_sha512_state_st_common_array ptr ({{ state.h }}, {{ state.sz }}, {{ state.block }}, {{ state.n }});
 };
 
 let pointer_to_fresh_sha512_state_st_array name = do {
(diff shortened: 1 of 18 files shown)


andreistefanescu:
The C code does not manipulate unbounded memory or symbolic size memory explicitly, that is, does not contain any loops. It does that implicitly though, through the use of `memcpy` and `memset`. In order to reason about the C code automatically, via SMT solvers, it is possible to encode similar operations on SMT arrays by using universal quantification, namely `arrayCopy` and `arraySet`. Similarly, it is possible to encode `arrayRengeEqual`, which specifies that two arrays are equal only on a given range. Specifically
```
forall i . arrayCopy(dest_array, dest_index, src_array, src_index, length)[i] =
  if i <= dest_index && i < dest_index + length then src_array[i - dest_index + src_index] else dest_array[i]
forall i . arraySet(dest_array, dest_index, value, length)[i] =
  if i <= dest_index && i < dest_index + length then value else dest_array[i]
arrayRange(dest_array, dest_index, src_array, src_index, length) = forall i j .
  i <= dest_index && i < dest_index + length && j <= src_index && j < src_index + length && i - j == dest_index - src_index =>
    dest_array[i] == src_array[j]
```
Z3 is able to prove the queries resulted from the verification process (with default settings). CVC4 returns unknown/timeout, while I don't know if Yices supports universal quantifiers over arrays.

andreistefanescu:
The x86_64 code contains a loop that updates the SHA internal state (H) for one block of input at a time. In order to reason about the x86_64 code automatically, we propose to implement a simple Abstract Interpretation style analysis that tracks the registers and the memory regions. Each register or memory location can be either a constant or a variable. The analysis also tracts inequalities between registers. Using the results of this analysis, it becomes possible to prove that all memory accesses are in bound, and then to extract a SAW term that captures the functional behaviors of the x86_64 function (as a fixpoint). Next, it's possible to use rewriting on the body of the extracted fixpoint to prove it's equivalent to the Cryptol specification.

SAW/proof/common/memory.saw
```diff
@@ -14,7 +14,7 @@
 /*
  * Specification of OPENSSL memory management functions
  */
-let OPENSSL_malloc_spec = do {
+let OPENSSL_malloc_spec is_init = do {
```
apetcher-amazon, line 17: Please add some comments about this parameter and how the value affects the spec.

SAW/proof/SHA512/SHA512-common.saw
```diff
@@ -216,18 +308,41 @@ include "memory.saw";
...
 
-enable_what4_hash_consing;
-
 // Verify the block data function assembly satisfies the
 // `sha512_block_data_order_spec` specification
-sha512_block_data_order_spec <- crucible_llvm_verify_x86 m "../../build/x86/crypto/crypto_test" "sha512_block_data_order"
+
+enable_what4_hash_consing;
+
+//sha512_block_data_order_ov <- llvm_verify_x86 m "../../build/x86/crypto/crypto_test" "sha512_block_data_order"
```
apetcher-amazon, line 316: For now, I think we should keep the bounded specs/proofs in place, since we are relying on them for the HMAC proof. We can remove these when we develop an unbounded HMAC proof.

SAW/proof/SHA512/SHA512.saw
```diff
@@ -5,141 +5,57 @@
...
+include "goal-rewrites.saw";
+
 // Verify the `EVP_SHA_INIT` C function satisfies the `EVP_sha_init_spec`
 // specification
 crucible_llvm_verify m EVP_SHA_INIT [] true EVP_sha_init_spec (w4_unint_yices []);
 
 // Verify the `EVP_DigestInit` C function satisfies the `EVP_DigestInit_spec`
 // specification
 crucible_llvm_verify m "EVP_DigestInit"
-  [ OPENSSL_malloc_ov
```
apetcher-amazon, line 15: Leave the bounded proof in place, for now.

apetcher-amazon (commented):
In addition to the other comments:

- There are a few lines of commented-out code here and there. Please remove them.
- Merge in the latest from the master branch to resolve conflicts

SAW/proof/SHA512/SHA512-common.saw
```diff
@@ -174,6 +204,71 @@ let pointer_to_fresh_sha512_state_st name n = do {
...
+ * which indicates that the field `name` of the struct pointed to by `ptr`
+ * contains the value of the global variable `GLOBAL`.
+ * All statements that do not match these two forms are documented inline
+ */
+// Specify the sha512_state_st_array struct from a SHAState
+let points_to_sha512_state_st_common_array ptr (h, sz, block, n) = do {
+  crucible_points_to_array_prefix (crucible_field ptr "p") block {{ `SHA512_CBLOCK : [64] }};
+
+  // Specify `sha512_state_st.Nl` and `sha512_state_st.Nh` contain `sz`
+  //llvm_points_to_at_type (crucible_field ptr "Nl") i128 (crucible_term sz);
```
apetcher-amazon, line 238: Remove commented out code

SAW/proof/SHA512/SHA512-common.saw
```diff
@@ -216,11 +311,12 @@ include "memory.saw";
...
 
-enable_what4_hash_consing;
-
 // Verify the block data function assembly satisfies the
 // `sha512_block_data_order_spec` specification
-sha512_block_data_order_spec <- crucible_llvm_verify_x86 m "../../build/x86/crypto/crypto_test" "sha512_block_data_order"
+
+enable_what4_hash_consing;
+
+sha512_block_data_order_spec <- llvm_verify_x86 m "../../build/x86/crypto/crypto_test" "sha512_block_data_order"
```
apetcher-amazon, line 319: Should be named sha512_block_data_order_ov?

SAW/proof/SHA512/goal-rewrites.saw
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+*/
+
+/*
+ * Rewrite rules proofs and goal tactics. This does not
+ * contain any specifications or assumtions, thus it does
```
apetcher-amazon, line 8: typo

apetcher-amazon (commented):
Also ensure comments from my last review are addressed.

andreistefanescu:
@apetcher-amazon I think I address all comments. We should update the readme to reflect this changes. Do you want to do that, or should I?

apetcher-amazon:
Can you update the README files (top level and SAW)? I think we need the following changes:

1. To keep the documentation simple, I suggest we just have one entry for SHA-2. So remove the 512 variant and the EVP_Digest operation from the list of things that are verified. 
2. Remove the InputLength caveat from SHA-2, and add any new caveats. Make sure the new assumption about memory being initialized is described (probably in a new caveat).
3. In the SAW README, update the informal description of the SHA-2 specs. Remove the table entry for EVP_Digest.

apetcher-amazon (approved):

PR #50, opened 2021-06-23
RSA verification issues with BN_is_odd

samuel40791765:
New RSA code in latest BoringSSL merge has a new `BN_is_odd` check on the `e` value of rsa. This slight change has caused our proof to fail, so we're going to need new ways to workaround the new `BN_is_odd` function that was introduced and the value of `e` to be odd.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,7 @@
 [submodule "src"]
 	path = src
-	url = https://github.com/samuel40791765/aws-lc.git
-	branch = boringssl-merge-7fffa4636cf7647daf981914286d5d32f1beab6d
+	branch = main
+	url = https://github.com/awslabs/aws-lc.git
 [submodule "cryptol-specs"]
 	path = cryptol-specs
 	branch = sha-imperative
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,7 @@
 [submodule "src"]
 	path = src
-	url = https://github.com/awslabs/aws-lc.git
-	branch = boringmerge20210317
+	url = https://github.com/samuel40791765/aws-lc.git
+	branch = boringssl-merge-7fffa4636cf7647daf981914286d5d32f1beab6d
 [submodule "cryptol-specs"]
 	path = cryptol-specs
 	branch = sha-imperative
--- a/SAW/proof/RSA/BN.saw
+++ b/SAW/proof/RSA/BN.saw
@@ -28,7 +28,7 @@ let BN_is_odd_spec num = do {
 
   crucible_execute_func [bn_ptr];
 
-  crucible_return (crucible_term {{ if integerFromBV bn % 2 == 1 then (1 : [32]) else 0 }});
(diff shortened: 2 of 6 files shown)


SAW/proof/RSA/RSA.saw
```diff
@@ -514,6 +515,7 @@ RSA_verify_PKCS1_PSS_mgf1_ov <- llvm_verify
   , EVP_DigestFinal_ex_0x68_ov
   , sha512_block_data_order_spec
   , BN_num_bits_n_bits_ov
+  , BN_is_odd_ov
```
apetcher-amazon, line 518: This override needs to go in the verify command for RSA_verify_pss_mgf1 near line 650, not here. This is just the padding function, which doesn't test for oddness.

SAW/proof/RSA/BN.saw
```diff
@@ -157,3 +169,13 @@ BN_num_bits_e_bits_ov <- llvm_verify m "BN_num_bits"
...
+// BN_is_odd_ov <- llvm_verify m "BN_is_odd"
+//  []
+//  true
+//  (BN_is_odd_spec n_words)
+//  (w4_unint_z3 []);
+
+BN_is_odd_ov <- crucible_llvm_unsafe_assume_spec
+  m
+  "BN_is_odd"
+  (BN_is_odd_spec n_words);
```
apetcher-amazon, line 181: This needs to be 1 instead of n_words. To avoid magic numbers, make a constant called e_words that has the value 1.

SAW/proof/RSA/BN.saw
```diff
@@ -19,6 +19,18 @@ let BN_num_bits_spec num_bits = do {
...
+let BN_is_odd_spec num = do {
+  let num_bits = eval_size {| num * 64 |};
+
+  bn_ptr <- crucible_alloc_readonly (llvm_struct "struct.bignum_st");
+  (bn, bn_d_ptr) <- ptr_to_fresh_readonly "bn" (llvm_int num_bits);
+  points_to_bignum_st_same bn_ptr bn_d_ptr num;
+
+  crucible_execute_func [bn_ptr];
+
+  crucible_return (crucible_term {{ if integerFromBV bn % 2 == 1 then (1 : [32]) else 0 }});
```
apetcher-amazon, line 31: parenthesis would make this more readable: if (integerFromBV bn) % 2 == 1 then

.gitmodules
```diff
@@ -1,7 +1,7 @@
 [submodule "src"]
 	path = src
-	url = https://github.com/awslabs/aws-lc.git
-	branch = boringmerge20210317
+	url = https://github.com/samuel40791765/aws-lc.git
```
apetcher-amazon, line 3: We can go ahead and merge this in with the assumption on BN_is_odd. Just update the submodule so it points to the latest on the aws-lc mainline. That we we know this change won't break anything in CI.

samuel40791765: Updated 👍

apetcher-amazon (approved):

Issue #49, opened 2021-06-09
Verification does not examine release mode


apetcher-amazon:
The build used by the SAW verification has -DCMAKE_BUILD_TYPE=Rel instead of -DCMAKE_BUILD_TYPE=Release. This results in the default build being verified instead of the release build.

RyanGlScott:
See #49, which switches CMake to use `Release` settings for all proofs except RSA.

apetcher-amazon:
Addressed by #51

PR #51, opened 2021-07-14
Use CMake's Release settings in most proofs

RyanGlScott:
This ports the proofs of SHA-2, HMAC, AES-GCM, AES-KW(P), EVP_*, ECDSA, and ECDH to use `-DCMAKE_BUILD_TYPE=Release`, fixing most of #49. Previously, it was using `=Rel`, which inadvertently caused the proofs to use the same optimization settings as if `=Debug` were used. The only proofs that this does not affect are the RSA proofs, which are left as future work.

With higher optimization settings, Clang more aggressively inlines certain functions. As a result, some functions can no longer be specified (e.g., `constant_time_select_w`), as they no longer exist in the compiled bitcode. This poses a slight challenge for the proofs in this repo, as several functions rely on using overrides for functions that are now optimized away. There are two general strategies that this patch uses to accommodate the additional inlining:

* Some functions are inlined down to invocations of very similar functions. For example, `ec_scalar_add` is now inlined away, revealing a call site to `bn_mod_add_words`, which is almost identical to `ec_scalar_add`. As a result, one adapt the specification for `ec_scalar_add` to use on `bn_mod_add_words` and override the latter instead of the former.

  Note that many of these adapted overrides, such as that of `bn_mod_add_words` are proven with `llvm_unsafe_assume_spec` because the original specifications that they were based on also use `llvm_unsafe_assume_spec`. As a result, this does not introduce _new_ unverified specifications, as these specifications are simply re-occurrences of existing unverified specifications.
* There are a handful of functions that SAW currently cannot handle when they are inlined away. For these functions, we opt to patch the `aws-lc` source code to mark them as `noinline` to prevent Clang from inlining them. Each of these functions are described in the top-level `README`, along with the algorithms that they are used in and explanations for why we opt to mark them as `noinline`.

Besides inlining, there are a variety of other minor issues posed by the increased optimization levels:

* When optimized, the bitcode for `bn_select_words` compares pointers from different allocation blocks. This is something which SAW disallows by default as a sanity check, but it can be disabled using the `enable_lax_pointer_ordering` command. This command is currently only available in recent nightly builds of SAW, so I updated the CI to use a more recent SAW version.
* Clang's argument promotion optimizations will change the type of `aes_gcm_from_cipher_ctx`'s argument from `struct evp_cipher_ctx_st*` to  `struct EVP_AES_GCM_CTX*`, which requires minor changes to `aes_gcm_from_cipher_ctx`'s specification to accommodate.
* The proof goals for three functions (`bn_reduce_once_in_place`, `ECDSA_do_sign`, and `ECDSA_do_verify`) change as a result of optimization, which requires changes to their proof scripts. In particular, these proof scripts now make use of the `hoist_ifs_in_goal` tactic, which is only available in recent nightly builds of SAW.

-----


SAW/proof/BN/BN.saw
```diff
@@ -165,7 +165,7 @@ let pointer_to_bn_mont_ctx_st num N = do {
   crucible_points_to N_d_ptr (crucible_term {{ (fromInteger `N) : [num_bits] }});
   points_to_bignum_st (crucible_field ptr "N") N_d_ptr num num {{ 0 : [32] }};
 
-  return ptr;
+  return (RR_d_ptr, ptr);
```
RyanGlScott, line 168: I had to modify the return value of this `pointer_to_bn_mont_ctx_st` utility in support of the new `bn_mod_mul_montgomery_small_RR_d_{,same}_spec` specifications, which require use of `RR_d_ptr`.

SAW/proof/EC/EC.saw
```diff
@@ -511,6 +707,20 @@ let ec_get_x_coordinate_as_bytes_spec = do {
   crucible_return (crucible_term {{ 1 : [32] }});
 };
 
+// This spec is currently a bit simplistic, as it will always return 1.
+// ec_GFp_simple_is_on_curve is used to check for fault attacks or bugs, and
+// as a result, it will never return a non-1 result unless something goes
+// unusually wrong.
+let ec_GFp_simple_is_on_curve_spec = do {
```
RyanGlScott, line 714: As noted in the comments above, this specification is somewhat simplistic. I've wondered if it would be worth trying to make the specification more precise and actually check the elliptic curve–related math it claims to check, but I have not investigated doing so yet, since in a normal use of `aws-lc`, this function will always return `1`.

If we decide to leave this specification as-is, do you think it is worth mentioning this caveat in the top-level `README`?

SAW/proof/ECDSA/ECDSA.saw
```diff
@@ -211,26 +200,28 @@ jacobian_affine_1_thm <- prove_print
 ECDSA_do_sign_ov <- llvm_verify m "ECDSA_do_sign"
   [ sha512_block_data_order_spec
   , ec_bignum_to_scalar_ov
-  , ec_scalar_to_montgomery_ov
-  , ec_scalar_to_montgomery_same_ov
-  , ec_scalar_from_montgomery_ov
-  , ec_scalar_from_montgomery_same_ov
-  , ec_scalar_add_ov
```
RyanGlScott, line 218: In this patch, I only removed overrides for functions that had direct counterparts (e.g., I removed `ec_scalar_add_ov` in favor of `bn_mod_add_words_ov`). It's likely that there are other overrides which are not actually used in the proofs, but I opted not to change the status quo in this regard. Do you think it would be worthwhile to go through and remove all such overrides that aren't used? (Note that many of them are unused regardless of optimization level.)

SAW/proof/ECDSA/goal-rewrites.saw
```diff
@@ -132,11 +132,51 @@ let mont_thms =
...
+  {{ \(x : [384]) -> fromInteger (toInteger x) == x }};
+eq_modulo_arith_thm <- prove_folding_theorem_w4
+  {{ \(x : Integer) (y : Integer) (z : Integer) ->
+    ((x * ((y - `P384_n) + z)) % `P384_n) == ((x * (y + z)) % `P384_n) }};
 
+// These theorems are true, but every SMT solver I've tried takes a really long time to prove them.
+// (It's possible to prove a similar theorem on a bitvector of a smaller length, such as 8.)
+// The uses of `apply` are to ensure that rewriting does not go into an infinite loop.
+let intLe = parse_core "\\(x y : Integer) -> intLe x y";
+toInteger_sub_thm_1 <- assume_folding_theorem
```
RyanGlScott, line 175: `toInteger_sub_thm_{1,2}` are the only uses of `assume_folding_theorem`—that is, the only uses of `prove_print assume_unsat`—in this patch. As the comments above indicate, I strongly believe that these theorems are true, but I was unable to convince any SMT solvers of this fact (at least, I couldn't make it work on length-384 bitvectors). Do you happen to know of any SAW tricks that might allow these proofs to go through?

RyanGlScott (commented):
I've left a handful of comments describing changes that warrant a closer look.

RyanGlScott:
I'd be interested in seeing the result of CI before removing the draft status of this PR. Can someone with the necessary permissions do so?

RyanGlScott:
Oops, I forgot to check in a patch used in the proofs. Could someone start the CI again? (Also, apologies for having to repeatedly ping for CI purposes.)

RyanGlScott:
OK, CI made it a little further that time. I've pushed something that I think will address the ECDH CI failure (approval pending).

RyanGlScott:
The timeout on the most recent CI run is a bit curious, and I don't think it's the fault of the PR, since I can the entire script locally without issues. Perhaps it's a transient failure that would disappear if the CI were restarted. (Alternatively, I'd be fine with waiting to address any review comments first.)

apetcher-amazon:
I restarted the job to see if it will complete this time. Note that the Github runner only has 7GB of RAM. When you are testing locally, make sure your container has the same amount of RAM.

SAW/proof/ECDH/evp-function-specs.saw
```diff
@@ -83,6 +83,7 @@ let EVP_PKEY_paramgen_spec = do {
 let EVP_PKEY_keygen_spec = do {
   d <- crucible_fresh_cryptol_var "d" {| Integer |};
   crucible_ghost_value ec_random_nonzero_scalar_out d;
+  crucible_ghost_value bn_rand_range_words_out      d;
```
andreistefanescu, line 86: is `ec_random_nonzero_scalar_out` unused now?

SAW/proof/EC/EC.saw
```diff
@@ -485,6 +666,21 @@ let ec_cmp_x_coordinate_spec = do {
   crucible_return (crucible_term {{ if (fromJacobian (jacobianFromMontBV p)).x % `P384_n == (scalarFromBV r) % `P384_n then 1 : [32] else 0 }});
 };
 
+// This spec is very similar to that of ec_cmp_x_coordinate, as that
+// function directly invokes ec_GFp_mont_cmp_x_coordinate. When compiled with
+// optimizations, ec_cmp_x_coordinate will be inlined away, exposing a
+// call site to ec_GFp_mont_cmp_x_coordinate.
+let ec_GFp_mont_cmp_x_coordinate_spec = do {
```
andreistefanescu, line 673: if `ec_cmp_x_coordinate_spec ` and `ec_GFp_mont_cmp_x_coordinate_spec ` are identical, let's keep only one. Different functions (with different names) can have the same specs.

RyanGlScott:
It timed out again, but strangely enough, it timed out in a _different_ place. What's strange is that the previous commit passed, despite there not being much difference between the two. I've tried testing both commits locally with `docker run --memory 7g`, but they both fail consistently around here:

<details>

```
[17:43:24.819] Verifying HMAC_Init_ex ...
[17:43:24.826] Simulating HMAC_Init_ex ...
[17:43:24.826] Registering overrides for `OPENSSL_cleanse`
[17:43:24.827]   variant `Symbol "OPENSSL_cleanse"`
[17:43:24.827] Registering overrides for `OPENSSL_free`
[17:43:24.827]   variant `Symbol "OPENSSL_free"`
[17:43:24.827] Registering overrides for `OPENSSL_malloc`
[17:43:24.827]   variant `Symbol "OPENSSL_malloc"`
[17:43:24.827] Registering overrides for `sha512_block_data_order`
[17:43:24.827]   variant `Symbol "sha512_block_data_order"`
[17:43:24.834] Matching 1 overrides of  OPENSSL_malloc ...
[17:43:24.836] Branching on 1 override variants of OPENSSL_malloc ...
[17:43:24.836] Applied override! OPENSSL_malloc
[17:43:24.836] Matching 2 overrides of  OPENSSL_free ...
[17:43:24.836] Branching on 1 override variants of OPENSSL_free ...
[17:43:24.836] Applied override! OPENSSL_free
[17:43:24.839] Matching 1 overrides of  sha512_block_data_order ...
[17:43:24.844] Branching on 1 override variants of sha512_block_data_order ...
[17:43:24.846] Applied override! sha512_block_data_order
[17:43:24.849] Matching 1 overrides of  sha512_block_data_order ...
[17:43:24.854] Branching on 1 override variants of sha512_block_data_order ...
[17:43:24.855] Applied override! sha512_block_data_order
[17:43:24.857] Matching 1 overrides of  OPENSSL_cleanse ...
[17:43:24.857] Branching on 1 override variants of OPENSSL_cleanse ...
[17:43:24.857] Applied override! OPENSSL_cleanse
[17:43:24.862] Matching 1 overrides of  OPENSSL_malloc ...
[17:43:24.863] Branching on 1 override variants of OPENSSL_malloc ...
[17:43:24.863] Applied override! OPENSSL_malloc
[17:43:24.863] Matching 2 overrides of  OPENSSL_free ...
[17:43:24.864] Branching on 1 override variants of OPENSSL_free ...
[17:43:24.864] Applied override! OPENSSL_free
[17:43:24.865] Matching 1 overrides of  sha512_block_data_order ...
[17:43:24.869] Branching on 1 override variants of sha512_block_data_order ...
[17:43:24.871] Applied override! sha512_block_data_order
[17:43:24.873] Matching 1 overrides of  OPENSSL_malloc ...
[17:43:24.875] Branching on 1 override variants of OPENSSL_malloc ...
[17:43:24.875] Applied override! OPENSSL_malloc
[17:43:24.875] Matching 2 overrides of  OPENSSL_free ...
[17:43:24.875] Branching on 1 override variants of OPENSSL_free ...
[17:43:24.875] Applied override! OPENSSL_free
[17:43:24.876] Matching 1 overrides of  sha512_block_data_order ...
[17:43:24.880] Branching on 1 override variants of sha512_block_data_order ...
[17:43:24.881] Applied override! sha512_block_data_order
[17:43:24.883] Matching 2 overrides of  OPENSSL_free ...
[17:43:24.883] Branching on 1 override variants of OPENSSL_free ...
[17:43:24.883] Applied override! OPENSSL_free
[17:43:24.917] Checking proof obligations HMAC_Init_ex ...
Calling Yices to check sat
Running check sat
Calling Yices to check sat
Running check sat
Calling Yices to check sat
Running check sat
Calling Yices to check sat
Running check sat
Calling Yices to check sat
Running check sat
2021/07/23 17:44:07 Failed to run saw script %s.verify-HMAC-SHA384-quickcheck.sawsignal: killed
exit status 1
```
</details>

I'm having a hard time imagining what would cause the apparent differences on a GitHub Actions runner.

SAW/proof/EC/EC.saw
andreistefanescu: I don't think it's worth making the spec precise and checking it

SAW/proof/ECDSA/ECDSA.saw
andreistefanescu: I prefer minimal changes, and I think removing unused overrides is outside the scope of this pull request. That could be a follow-up pull-request, but I don't think critical

SAW/proof/ECDSA/goal-rewrites.saw
andreistefanescu: I'm thinking let's check for 8 bits, and assume for 384. The issue with `toInteger` is that SAW bit-blasts it, before it get to the SMT solver, and at that point there's nothing that can fix it. I've thinking about fixing this by translating `toInteger` to `bv2int` in Z3 (which is just in Z3, not SMTLib), but I think it's out of scope here

SAW/proof/BN/BN.saw
andreistefanescu: either this, or add an additional helper that drops `RR_d_ptr`

andreistefanescu (commented):
Looks good! See the inline comments.

SAW/proof/EC/EC.saw
RyanGlScott: Sounds good to me.

SAW/proof/ECDH/evp-function-specs.saw
RyanGlScott: It's unused in the sense that the only place where it is used is `ec_random_nonzero_scalar_spec`, but that spec isn't actually applied anyway. I generally erred on the side of not removing existing specs if I didn't absolutely have to, but I could remove it if you'd prefer.

SAW/proof/ECDSA/ECDSA.saw
RyanGlScott: Sounds good to me.

SAW/proof/ECDSA/goal-rewrites.saw
RyanGlScott: I like the idea of proving it for 8 bits. Unfortunately, I can't simply copy-and-paste the existing theorems and replace `384` with `8`, since `P384_n` won't fit in 8 bits. Still, given that the 8-bit versions of the theorem are just proofs-of-concept, I opted to replace `P384_n` with a morally equivalent 8-bit value to show that the same idea is feasible at a smaller bit width.

SAW/proof/common/helpers.saw
```diff
@@ -69,5 +69,7 @@ let llvm_verify_x86 m elf func globals pathsat spec tactic = if do_prove
   then crucible_llvm_verify_x86 m elf func globals pathsat spec tactic
   else crucible_llvm_unsafe_assume_spec m func spec;
 
-let prove_folding_theorem t = prove_print abc (rewrite (cryptol_ss ()) t);
+let prove_folding_theorem t = prove_print z3 (rewrite (cryptol_ss ()) t);
```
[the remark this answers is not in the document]
RyanGlScott: Good idea. Done.

RyanGlScott:
Hooray, the most recent commit passed CI! I'll go ahead and remove the draft status.

Does anyone else have additional review comments? Also, given that there's quite a few commits in this PR, should I squash them all myself, or will the person who lands the PR squash the commits themselves?

README.md
```diff
@@ -32,11 +32,24 @@ The caveats associated with some of the verification results are defined in the
...
+| `value_barrier_w` | <nobr>AES-KW(P)</nobr>, ECDSA | This function directly uses inline assembly. Without `noinline`, this inline assembly would be inlined into several larger functions which cannot currently use `llvm_verify_x86`. Note that `value_barrier_w(x)` is semantically equivalent to `x` for all `x`; `value_barrier_w` primarily exists to prevent Clang from applying certain optimizations to `x` which would adversely affect constant-time code. We have verified a variation of `value_barrier_w` that _does_ get inlined using an experimental version of SAW, which we aim to contribute to upstream SAW soon. |
```
apetcher-amazon, line 52: Please remove the last sentence---I don't think we should put plans in the documentation like this. If you want to keep track of this work item, create an issue in the Github issue tracker for this repository.

apetcher-amazon (commented):
Nice work. The documentation is very clear. Just address the comment about the README and I'll approve. 

No need to squash commits. All PRs are squashed when they are merged.

apetcher-amazon:
Waiting for a good workflow run, and also a check to make sure this PR won't break the AWS-LC build. If all goes well, I should be able to approve and merge tomorrow.

bryce-shang:
This PR has been tested with latest aws-lc https://github.com/awslabs/aws-lc-verification/pull/52

apetcher-amazon:
Still waiting on a complete test of all cases. Should be done in the next day or so. Sorry for the delay.

apetcher-amazon:
My test is running out of memory in the GCM proof when the input length is 10. I'm still investigating, but do you mind trying this case locally (in a container with 7GB of RAM) and see if it works for you? The easiest way to do this is to modify verify-AES-GCM-quickcheck.saw by setting evp_cipher_update_len to 10.

RyanGlScott:
Ah, my apologies—I wasn't aware there was a memory requirement in place. If I run the proofs in a Docker container with `--memory 7g`, then I am indeed able to reproduce the out-of-memory issue. I suspect that Yices is using more memory on the optimized proofs, which _just_ barely increases the peak memory usage over 7 GB. If I increase the limit to `--memory 8g`, then I can complete the proofs. Is that an acceptable increase?

bryce-shang:
> Ah, my apologies—I wasn't aware there was a memory requirement in place. If I run the proofs in a Docker container with `--memory 7g`, then I am indeed able to reproduce the out-of-memory issue. I suspect that Yices is using more memory on the optimized proofs, which _just_ barely increases the peak memory usage over 7 GB. If I increase the limit to `--memory 8g`, then I can complete a run of `verify-AES-GCM-quickcheck.saw`. Is that an acceptable increase?

Yes, increase the memory in https://github.com/awslabs/aws-lc-verification/pull/53

apetcher-amazon:
I think increasing the memory available solves the problem, for now. We will slow down the CI a bit to give more memory per process. No need to optimize the proofs unless the cases we check in the Github runner (what has a hard limit of 7GB) start failing.

apetcher-amazon:
Waiting on integration of #53 before I approve.

apetcher-amazon (approved):

PR #58, opened 2022-03-09
Proof for optimized HMAC code.

apetcher-amazon:

This is the proof for the [optimized HMAC code](https://github.com/awslabs/aws-lc/pull/368) which changed the HMAC context structure and the code that uses it.


SAW/proof/HMAC/HMAC.saw
```diff
@@ -5,17 +5,192 @@
...
+  points_to_hmac_in_place_methods (crucible_elem (crucible_global HMAC_METHODS_STORAGE) 0);
+};
+
+llvm_verify m HMAC_METHODS_INIT
+  evp_methods_ovs
+  true
+  hmac_in_place_methods_init_spec
+  (w4_unint_z3 []);
+
+hmac_in_place_methods_ov <- llvm_unsafe_assume_spec
```
robdockins, line 160: Similar, this looks like a new assumption.

SAW/proof/HMAC/HMAC.saw
```diff
@@ -5,17 +5,192 @@
...
+  llvm_verify m (md_init index)
+    []
+    true
+    (evp_md_methods_init_spec index)
+    (w4_unint_z3 []);
+};
+
+for MD_INDICES evp_methods_init_ov;
+
+let evp_methods_ov index = do {
```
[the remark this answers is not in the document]
apetcher-amazon: Added comments

SAW/proof/HMAC/HMAC.saw
apetcher-amazon: Added comments.

SalusaSecondus:
I cannot provide meaningful review of the proof since I don't understand it.

However, do we prove that reusing an `HMAC_CTX` for multiple HMAC calculations (without setting the key, so passing `NULL` in for both the `evpmd` and `key` fields) is correct?

apetcher-amazon:
> ... do we prove that reusing an `HMAC_CTX` for multiple HMAC calculations (without setting the key, so passing `NULL` in for both the `evpmd` and `key` fields) is correct?

No. We only prove HMAC correct when the `HMAC_CTX` passed to `HMAC_Init_ex` or `HMAC` is in its zeroized state. See the [SAW Readme](https://github.com/apetcher-amazon/aws-lc-verification/blob/master/SAW/README.md) for an informal description of what is verified.

SalusaSecondus:
> > ... do we prove that reusing an `HMAC_CTX` for multiple HMAC calculations (without setting the key, so passing `NULL` in for both the `evpmd` and `key` fields) is correct?
> 
> No. We only prove HMAC correct when the `HMAC_CTX` passed to `HMAC_Init_ex` or `HMAC` is in its zeroized state. See the [SAW Readme](https://github.com/apetcher-amazon/aws-lc-verification/blob/master/SAW/README.md) for an informal description of what is verified.

Is it reasonable/possible to extend the proof to cover this case?

apetcher-amazon:
> 



> > > ... do we prove that reusing an `HMAC_CTX` for multiple HMAC calculations (without setting the key, so passing `NULL` in for both the `evpmd` and `key` fields) is correct?
> > 
> > 
> > No. We only prove HMAC correct when the `HMAC_CTX` passed to `HMAC_Init_ex` or `HMAC` is in its zeroized state. See the [SAW Readme](https://github.com/apetcher-amazon/aws-lc-verification/blob/master/SAW/README.md) for an informal description of what is verified.
> 
> Is it reasonable/possible to extend the proof to cover this case?

Yes, this would be a relatively simple enhancement, and it would be a good exercise for someone who is just learning how to use SAW. Feel free to create a ticket in the Github issue tracker to track this feature request, or submit a PR.

apetcher-amazon:
While testing this change, I found a performance bug in the HMAC proof script that was used in the full CI run. This script was testing all key sizes up to 129 bytes. This was not intentional, and was causing the full check to take too long. I changed it so that the full check doe the same thing as quick check---verify one "short" key (128 bytes) and one "long" key (129 bytes). We could add more key sizes here, but the long-term plan is to replace this with an unbounded proof, which will cover all key sizes.

andrewhop (approved):

PR #65, opened 2022-08-16
Remove some field names from structs

apetcher-amazon:
A recent [change](https://github.com/awslabs/aws-lc/pull/590) to the upstream repo of AWS-LC caused the compiler to remove some field names from structs in the optimized build. This change removes the references to those names from the proofs and uses offsets instead. 

### DIFF
--- a/SAW/proof/EVP/EVP_CTX.saw
+++ b/SAW/proof/EVP/EVP_CTX.saw
@@ -31,23 +31,40 @@ let points_to_evp_pkey_method_st
       derive
       paramgen
       ctrl = do {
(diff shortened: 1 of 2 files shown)


bryce-shang:
The direct cause is [crypto/fipsmodule/hkdf/hkdf_test.cc](https://github.com/awslabs/aws-lc/pull/590/commits/28ba57ba1f8335ca66d1a53e08fa4c6916fe9ef4#diff-d15a837a283e2c9047302a18e2745ecdb24eb0cca47636171182758145a3e714). Other diffs of the commit does not fail the proof.

andrewhop (approved):

bryce-shang:
`CryptoAlg-1267` is created to track the cause.

PR #74, opened 2022-10-20
Parallel the SAW proofs

pennyannn:

Notes:
1. Update runner pool to use ubuntu 16core
2. Add a Python script for paralleling SAW proofs
3. Reduce the proof output to only contain useful information. The complete output can be generated through debug mode.
4. Report run time

### DIFF
--- a/SAW/scripts/parallel.py
+++ b/SAW/scripts/parallel.py
@@ -1,0 +1,102 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+import argparse
+import multiprocessing
+import os
+import re
+import subprocess
+import sys
+import time
+
+LINEWIDTH = 80
+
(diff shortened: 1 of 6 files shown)


SAW/scripts/parallel.py
```diff
@@ -0,0 +1,102 @@
...
+    print("Finishing proof with command {}".format(command))
+    return [result, end - start]
+
+# Create a summary for the proof output
+def create_summary(output, error, exit_code, debug):
+    output = output.split('\n')
+    summary = []
+    for l in output:
+        if not debug:
+            if not l or any(regex.match(l) for regex in patterns): continue
```
andrewhop, line 56: How long does this take to run? Is it worth the effort to create one regex that matches on all the patterns?

pennyannn: Good idea. Check the new version.

andrewhop (approved):

PR #84, opened 2023-01-20
Fix memory usage issue in SHA512-384-SELECTCHECK

pennyannn:
When upgrading SAW binary from an older [version](https://github.com/pennyannn/saw-script/commits/old) to a newer [version](https://github.com/pennyannn/saw-script/commits/yppe/match-concrete-size-array-1), it causes a performance downgrade for verify-SHA512-384-selectcheck. Each selectcheck job used to take around 7-8G, and now 36-48G (this is tested on a 64G memory machine). It is highly possible that some changes in SAW caused this memory utilization issue. 

Profiling of the proof runs shows that we can use Haskell runtime control to limit each run to use only 30GB of memory (this causes the run time to increase a bit probably due to garbage collection). With this limit, each job takes at most 15mins. There are a total of 12 jobs. Given that the codebuild machine can provide 138GB of memory, theoretically, we can achieve parallelism of 4, and the total run time should be around 40-45 mins. An actual experiment on a machine with 72core and 144G memory results in 38min elapsed time.

This PR limits heap size of each job to be 30G, and dynamically calculate number of parallel processes based on size of free memory.

### DIFF
--- a/SAW/proof/AES/AES-GCM-check-entrypoint.go
+++ b/SAW/proof/AES/AES-GCM-check-entrypoint.go
@@ -38,7 +38,7 @@ func main() {
 		wg.Add(1)
 		saw_template := "verify-AES-GCM-selectcheck-template.txt"
 		placeholder_name := "TARGET_LEN_PLACEHOLDER"
-		go utility.CreateAndRunSawScript(saw_template, placeholder_name, i, &wg)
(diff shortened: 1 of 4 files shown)


andrewhop:
[Memory utilization issue in more recent SAW](https://github.com/GaloisInc/saw-script/issues/1805)

andrewhop (approved):

PR #88, opened 2023-01-26
Unset bits in ia32cap to disable AVX512[F|DQ|BW|VL] instructions

pennyannn:
This PR unsets bits in ia32cap to disable AVX512[F|DQ|BW|VL] instructions. The saw proof only verifies the AES-GCM code when the AVX512 instructions are disabled. Note that this only happens for AES-GCM, all other proofs still use an ia32cap that is fully set.

AWS-LC PR: https://github.com/awslabs/aws-lc/pull/692
Ticket: P80161838


README.md
```diff
@@ -10,7 +10,7 @@ AWS libcrypto includes many cryptographic algorithm implementations for several
 | ----------| -------------| --------------- | -----------| ------------ | --------- |
 | SHA-2     | 384, 512     | EVP_DigestInit, EVP_DigestUpdate, EVP_DigestFinal | SandyBridge+ | NoEngine, MemCorrect | [SAW](SAW/README.md) |
 | HMAC      | with <nobr>SHA-384</nobr> | HMAC_CTX_init, HMAC_Init_ex, HMAC_Update, HMAC_Final, HMAC | SandyBridge+ | NoEngine, MemCorrect, InitZero, CRYPTO_once_Correct | [SAW](SAW/README.md) |
-| AES-GCM   | 256 | EVP_CipherInit_ex, EVP_EncryptUpdate, EVP_DecryptUpdate, EVP_EncryptFinal_ex, EVP_DecryptFinal_ex | SandyBridge+ | InputLength, NoEngine, MemCorrect, InitZero, AESNI_GCM_Patch, AES_GCM_FROM_CIPHER_CTX_Correct, NoInline | [SAW](SAW/README.md) |
+| AES-GCM   | 256 | EVP_CipherInit_ex, EVP_EncryptUpdate, EVP_DecryptUpdate, EVP_EncryptFinal_ex, EVP_DecryptFinal_ex | SandyBridge+ | InputLength, NoEngine, MemCorrect, InitZero, AESNI_GCM_Patch, AES_GCM_FROM_CIPHER_CTX_Correct, NoInline, AES_GCM_AVX512 | [SAW](SAW/README.md) |
```
apetcher-amazon, line 13: The fact that the AVX-512 implementation is not verified shouldn't be listed as a caveat. Instead, we should define a new platform, add it to the list of platforms in the top-level README, and use this platform name in the row for AES-GCM. Remember to modify the AES-GCM row in both the top-level README and the SAW README.

Coming up with good platform names is hard, since the name has to refer to a large variety of microarchitectures and capabilities. Maybe something like `SandyBridge-Skylake`---since this is the range of microarchitectures supported by the proof. For the description of the platform, you can just copy the `SandyBridge+` description and add "but not AVX-512".

pennyannn: Done.

apetcher-amazon (approved):

PR #92, opened 2023-02-17
Update go verison used in docker file

andrewhop:
AWS-LC is working to update the version of Go used in our CI to unblock adoption of modern Go features. This change is inspired by how we install a particular version of go in https://github.com/aws/aws-lc/commit/6704d7f793fb95fb4421fcf78f74117df8d10a27.

I considered moving to a newer Ubuntu version which would come with a newer Go version but that would change the compiler as well (Clang 10 to 15) and I am afraid that would create issues with the formal verification.

A ENV command is a variable available during build time and runtime (which we need for goroot and path), while an ARG is only available during build time (and can be overwritten with a CLI parameter if you want to).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,9 +3,15 @@
 
 
 FROM ubuntu:20.04
+ENV GOROOT=/usr/local/go
+ENV PATH="$GOROOT/bin:$PATH"
+ARG GO_VERSION=1.20.1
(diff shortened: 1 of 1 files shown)


pennyannn (approved):
The CI passes meaning the aws-lc src code has been successfully compiled and the proofs  run successfully after this upgrade. The aws-lc CI runs additional selectchecks that are not tested in this CI. I have tested that the selectchecks still run successfully after this change. So I approve.

PR #98, opened 2023-04-12
Extend ECDSA proof to be unbounded

pennyannn:
This PR updates ECDSA proof to be unbounded. That is, the ECDSA functions are verified to be functionally correct with respect to arbitrary input message length (modulo the underlying EC operations). The set of top-level functions that are verified include `EVP_DigestSignInit`, `EVP_DigestVerifyInit`, `EVP_DigestSignUpdate`, `EVP_DigestVerifyUpdate`, `EVP_DigestSignFinal`, `EVP_DigestVerifyFinal`, `EVP_DigestSign` and `EVP_DigestVerify`.

In addition, this PR refactors the proofs in SHA512 directory and removes the selectcheck for SHA384.


samuel40791765 (commented):
Just reviewing the impact of the CI changes for AWS-LC:
The logic and the go script used for `SHA512_384_SELECTCHECK` was removed. This change shouldn't impact our CI to fail, but the existing `*_formal_verification_sha_selectcheck` will become a duplicate of 
`*_formal_verification_quickcheck`, since `SAW/scripts/run_checks_release.sh` will run the quick checks when no recognized `*_SELECT_CHECK` environment variable is found. We can delete the dimension once this PR is merged, but a coordinated merge shouldn't be necessary.
* https://github.com/aws/aws-lc/blob/main/tests/ci/cdk/cdk/codebuild/github_ci_linux_x86_omnibus.yaml#L369
* https://github.com/aws/aws-lc/blob/main/tests/ci/cdk/cdk/codebuild/github_ci_linux_x86_omnibus.yaml#L380

My only slight concern would be how much slower the CI becomes with the SHA512 proof being moved into the quick check. The current quick check is pretty fast (20 min), so there's plenty of leeway for us to move new proofs within it, but I just wanted to call this out so we're aware of it.

pennyannn:
> My only slight concern would be how much slower the CI becomes with the SHA512 proof being moved into the quick check. The current quick check is pretty fast (20 min), so there's plenty of leeway for us to move new proofs within it, but I just wanted to call this out so we're aware of it.

Thanks for the review! This PR doesn't add additional overhead to the existing quick check proof. What it does is that it replaces the existing ECDSA bounded proof with an unbounded proof. This unbounded proof runs in about the same time as the bounded version. Current quick check run still finishes in about 20 mins.

So essentially it should make the AWS-LC CI use less compute power (with the seleck check removed) without sacrificing speed.

SAW/spec/ECDSA/ECDSA.cry
```diff
@@ -0,0 +1,203 @@
...
+module ECDSA::ECDSA where
+
+import Array
+import Common::ByteArray
+import Primitive::Keyless::Hash::SHA384
+import HASH::SHA384
+import Common::ModDivZ
+
+// -------------------------------------------------------------------------
+// The following are copied from Common::EC::EC_P384 for resolving naming conflict
```
apetcher-amazon, line 15: Is copying everything the only option? Cryptol's module system appears offer some solutions to name collision problems. See `F.22`, `F.23`, and `F.24` of [this manual](https://cryptol.net/files/ProgrammingCryptol.pdf).

SAW/proof/SHA512/evp-function-specs.saw
```diff
@@ -129,6 +54,7 @@ let digestOut_post withLength out_ptr s_ptr out_value = do {
   };
 };
 
+// This bounded specification could not be removed because it is used by the bounded RSA proof
```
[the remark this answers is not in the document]
pennyannn: Good point. Done.

SAW/spec/ECDSA/ECDSA.cry
pennyannn: It seems like there is no good solution when importing Cryptol modules with naming collision into SAW. I submitted an [issue](https://github.com/GaloisInc/saw-script/issues/1857) to the SAW-script repository for feature request.

weaversa: SAW does support qualified naming of Cryptol modules using the `as` keyword combined with `import`. Here is an example -- Does this help?

```
$ cat fun.cry
f x = x + 1
$ saw
 ┏━━━┓━━━┓━┓━┓━┓
 ┃ ━━┓ ╻ ┃ ┃ ┃ ┃
 ┣━━ ┃ ╻ ┃┓ ╻ ┏┛
 ┗━━━┛━┛━┛┗━┛━┛ version 0.9.0.99 (<non-dev-build>)

sawscript> import "fun.cry" as fun
sawscript> print {{ fun::f 10 }}
[00:30:21.925] Assuming a = Integer
[00:30:21.925] 11
sawscript> :q
```

pennyannn: Thanks for the clarification. I tried a version where I put the quotation mark in the wrong place (Module::`w) and ended up getting an error. This lead to the misunderstanding that this feature is supported. Will be closing the issue. Sorry for the confusion it might have caused and thanks again!

apetcher-amazon (approved):

PR #112, opened 2023-07-21
Fix formal verification failure in AWS-LC PR 1101

pennyannn:
AWS-LC PR: https://github.com/aws/aws-lc/pull/1101
Ticket: V964947909

Changes from AWS-LC PR:
1. `EC_RAW_POINT` is renamed to `EC_JACOBIAN`.
2. A new parameter `Htable` is added to the `aesni_gcm_en[de]decrypt` functions. The assembly now directly refers to `Htable` instead of locating `Htable` with respect to `Xip` with some offset.


nebeid:
My understanding is that for point 1, `EC_JACOBIAN` change is accepted by the tests and they continue to pass.
But for point 2, there are changes in the proofs, but the AES-GCM tests were still failing so they were disabled. Is my understanding correct?

pennyannn:
> My understanding is that for point 1, `EC_JACOBIAN` change is accepted by the tests and they continue to pass. But for point 2, there are changes in the proofs, but the AES-GCM tests were still failing so they were disabled. Is my understanding correct?

Yes. Exactly right. The EC proofs now work with the new name `EC_JACOBIAN`. The AES-GCM proofs are updated to fit with the changes but they are currently crashing because of the lack of feature in SAW (Though I won't know for sure if the changes are correct until I am able to run the proofs. Maybe I can revert the changes back to what they used to be if you prefer. In principal it shouldn't matter because the proofs are disabled anyways). So I disabled them in the shell scripts.

SAW/scripts/release_jobs.sh
```diff
@@ -8,7 +8,9 @@ saw proof/SHA512/verify-SHA512-384.saw
 saw proof/SHA512/verify-SHA512-512.saw
 saw proof/HMAC/verify-HMAC.saw
 saw proof/KDF/verify-HKDF.saw
-cd proof/AES && go run AES-GCM-check-entrypoint.go
+# The proof for aesni_gcm_en[de]crypt requires having more than 6 function arguments.
+# Disabling the following proof due to lack of support of such a scenario in SAW.
+# cd proof/AES && go run AES-GCM-check-entrypoint.go
```
apetcher-amazon, line 13: Need to update the README files to indicate that AES-GCM is no longer formally verified.

pennyannn: Updated.

pennyannn: Done removing all entries related to AES-GCM.

apetcher-amazon (approved):

PR #139, opened 2024-01-23
Add unbounded AES-GCM verification.

andreistefanescu:
Caveats
* `llvm_verify_fixpoint_x86` proves *partial correctness*, that is, the property is verified just for terminating executions of the target x86_64 function
* induction proofs for `cryptol` functions are in fact proofs just for the inputs on which the function evaluation terminates, for example, `\x -> f x == g x` means that `f x` is equal with `g x` when `f x` terminates; moreover, the induction principle is applied manually, by using `prove_theorem assume_unsat` to create a rewrite rule that is then used in it's own proof, after the function `f` is unfolded.
* `saw` branch (with submodules): https://github.com/GaloisInc/saw-script/tree/sygus2

### DIFF
--- a/SAW/proof/AES/AES-CTR32.saw
+++ b/SAW/proof/AES/AES-CTR32.saw
@@ -1,0 +1,168 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+*/
+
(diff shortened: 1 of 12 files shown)


andreistefanescu:
I managed to run the entire proof as is, it takes 22 minutes on my M2 Max, memory usage peaked at 97% :-).

pennyannn:
Hi @andreistefanescu , which versions of CVC5 and Z3 were you using for running the entire proof?

PR #141, opened 2024-03-28
Improvements to formal verification of P-384

apetcher-amazon:

This PR has the following significant improvements to the formal verification of P-384
* Constant values like the pre-computed table of multiples of the base point are no longer assumed to be correct. Instead, these values are tested using formally-verified tests, and we assume these tests pass in the Coq proof. These assumptions are sufficient to prove that the constant values are correct.
* The ECDH output conversion, including the field inverse square operation and Jacobian to affine conversion, is formally verified. After this change, all field/point conversion operations are either verified in these proofs or provided by formally verified field arithmetic libraries. 

The README file was updated to remove the two significant caveats associated with the changes above. I also added two new caveats:
* ECDH_InfinityTestCorrect describes an assumption about the behavior of a particular subroutine used in ECDH output conversion. Previously, the entire conversion routine (not just this subroutine) was assumed to be correct.
* SAW_EC_FieldGap is a caveat that should have been included in the last PR on this topic. This caveat makes it clear that there is a gap between the semantics and memory model used to verify the field arithmetic vs other P-384 operations. 

There are also significant organizational changes to the Coq proof in this PR. These changes should improve the overall organization, and they were helpful in isolating Coq performance issues that arise from operating on actual values.


SAW/proof/EC/EC_P384_validate.saw
```diff
@@ -0,0 +1,141 @@
...
+p384_validate_base_table_ov <- llvm_verify m "p384_validate_base_table"
+  [ p384_validate_base_table_row_ov,
+	point_double_n_20_ov
+  ]
+  true
+  p384_validate_base_table_spec
+  (do {
+    w4_unint_z3 ["validate_base_row", "point_double_base_tsize"];
+  });
+
```
pennyannn, line 140: Should we also verify `p384_validate_one` and `p384_validate_constants` for completeness?

SAW/spec/EC/EC_mul.cry
```diff
@@ -305,3 +304,70 @@ lookup_table_fn p_table i acc j = if i == j then (p_table@j) else acc
...
+    res = (jp', r && r') 
+
+validate_base_table : [base_tsize][tsize]affine_point -> Bit
+validate_base_table t = res
+  where
+    ap = (t@0)@0
+    jp = [ap@0, ap@1, p384_felem_one]
+    (acc, res) = (foldl validate_base_table_body (jp, 1) t)
+
+ECDH_derive_bv : JacobianBVPoint -> scalar -> [bit_length]
```
pennyannn, line 368: Should this function be in EC_mul.cry or SAW/spec/ECDH/ECDH.cry (This file doesn't exist in the repo. It needs to be created) ?

README.md
```diff
@@ -78,6 +78,8 @@ The caveats associated with some of the verification results are defined in the
 | SAWBreakpoint | The proof uses SAW's breakpoint feature. This feature assumes the specification on the breakpoint function for the inductive hypothesis. The feature lacks well-foundedness check for the inductive invariant. |
 | ArmSpecGap | The Cryptol specification used in NSym proofs for Arm is different from the one used in the corresponding SAW proofs. Specifically, recursive comprehensions are written as recursions. We verify the body of the recursions are equivalent but the top-level loop structure is not verified. |
```
pennyannn, line 79: Now that we have so many tools and there will be gaps between every two of them, do you think if it is better to just have one item called `ToolGap` and put `ArmSpecGap`, `EC_Fiat_Crypto`, `SAWNSymGap`, and `SAW_EC_FieldGap` all underneath it? I would imagine `ToolGap` taking one row and in `Description`, we divide it into four rows for each.

pennyannn (commented):
I reviewed the SAW proof part of this PR.

SAW/spec/EC/EC_mul.cry
apetcher-amazon: I made a new file called `ECDH.cry`, but it is in `SAW/spec/EC`. I tried putting it in a new directory `SAW/spec/ECDH`, but that causes some problems for Cryptol imports that would probably require refactoring to fix. Putting this file in `SAW/spec/EC` seems like a good compromise.

README.md
apetcher-amazon: I agree that `SAW_EC_FieldGap` and `SAWNSymGap` are essentially the same (but for different pairs of tools), so I created `ToolGap` to cover them. The other caveats you mention are more about differences in functional specifications (rather than the semantics, etc. used by tools), so I think it makes sense to keep them separate.

Coq/EC/EC_P384_Abstract_5_equiv.v
```diff
@@ -0,0 +1,1111 @@
...
+
+Local Arguments cons [_] h [_] t.
+Local Arguments append [m] [n] [a]%type_scope {Inh_a} x y.
+Local Arguments bvOr [n] _ _.
+Local Arguments bvAnd [n] _ _.
+Local Arguments reverse [n] [a]%type_scope {Inh_a} _.
+Local Arguments bvSub [n] _ _.
+Local Arguments SAWCorePrelude.map [a]%type_scope {Inh_a} [b]%type_scope f%function_scope _ _.
+
+
```
pennyannn, line 50: Could you add a comment describing what this file contains?

apetcher-amazon: Done.

Coq/EC/ECEqProof.v
```diff
@@ -1042,4909 +445,473 @@ Section ECEqProof.
...
 
   Qed.
 
+
   End PointMulBase.
 
+  Variable felem_from_bytes : Vector.t bool 384 -> F.
+  Variable felem_to_bytes : F -> Vector.t bool 384.
+  Hypothesis felem_from_to_bytes_inv : forall x, felem_from_bytes (felem_to_bytes x) = x.
+  Hypothesis felem_to_from_bytes_inv : forall x,  felem_to_bytes (felem_from_bytes x) = x.
```
khieta, line 839: Add explanation for these variables and hypotheses

Coq/EC/ECEqProof.v
```diff
@@ -1042,4909 +445,473 @@ Section ECEqProof.
...
+
+  Definition jacobianToAffine_abstract x := @jacobianToAffine Fsquare Fmul felem_from_bytes felem_to_bytes x.
+  Definition to_bytes_3 (p : F * F * F) :=
+    (felem_to_bytes (fst (fst p)), (felem_to_bytes (snd (fst p)), felem_to_bytes (snd p))).
+  Definition from_bytes_2 p :=
+    (felem_from_bytes (fst p), felem_from_bytes (snd p)).
+
+  Variable F_order_correct : nat -> Prop.
+  Context `{F_FiniteField : @FiniteField F_order_correct F Feq _ _ _ _ _ _ _ _ F_field}.
+  Hypothesis p384_field_order_correct : F_order_correct p384_field_order.
```
khieta, line 849: Add explanation for these variables and hypotheses

Coq/EC/Util.v
```diff
@@ -290,6 +322,38 @@ Theorem even_of_nat_equiv : forall n,
 
 Qed.
 
+Theorem sub_eq_mono : forall x1 x2 y1 y2,
+  x1 = x2 ->
+  y1 = y2 ->
+  (x1 - y1)%nat = (x2 - y2)%nat.
```
khieta, line 328: Several of the lemmas in this file seem like they could be discharged by builtin tactics like `lia`. Are you separating them out for performance reasons?

Coq/EC/EC_P384_Abstract_Model_equiv.v
```diff
@@ -0,0 +1,4680 @@
...
+    jac_eq (fromPoint a) (seqToProd a') ->
+    jac_eq (fromPoint b) (seqToProd b') -> 
+    jac_eq (fromPoint (Curve.add a b)) (seqToProd (point_add true a' b')).
+
+  Hypothesis point_double_minus_3_jac_eq : forall (p:Curve.point) p',
+      jac_eq (fromPoint p) (seqToProd p') ->
+      jac_eq (fromPoint (Curve.double_minus_3 p)) (seqToProd (point_double p')).
+
+  Hypothesis sbvToInt_sign_extend_16_64_equiv : forall x,
+    sbvToInt _ (sign_extend_16_64 x) = sbvToInt _ x.
```
khieta, line 731: Add explanation for these variables and hypotheses

(Similar comment for other uses of `Hypothesis` in this file, but wanted to avoid spamming comments.)

khieta (approved):
I looked through the files in `Coq/EC/` and the README and didn't find anything problematic -- the Coq changes are largely refactoring.

Some more high-level comments:

1. I left a couple minor comments about documenting assumptions in the proofs when a comment wasn't readily available. For easier review in the future, you might consider adding a special comment string (e.g. "ASSUMPTION") to [assumptions](https://coq.inria.fr/doc/V8.18.0/refman/language/core/assumptions.html#assumptions) like `Variable` and `Hypothesis`.
If you're in the mood for some hacking, you could even set up GitHub CI to check that all assumptions are documented, and nothing is admitted. We do something sort of similar in [our Rust code](https://github.com/cedar-policy/cedar/blob/main/panic_safety.sh), where we check that any function that can panic is prefixed by a comment with the string "PANIC SAFETY".

2. How are the new table validation functions actually run? I see that they're assumed to pass in the proofs, but I didn't notice them being extracted anywhere.

SAW/scripts/x86_64/entrypoint_check.sh
```diff
@@ -16,8 +16,22 @@ apply_patch() {
 
 go env -w GOPROXY=direct
 
-# First, apply some patches (TODO: remove them)...
+ # Start by patching in the constant validation tests and executing them
+apply_patch "p384_validate"
```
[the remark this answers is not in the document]
apetcher-amazon: The patches are mostly used to make trivial modifications to the code (that we believe produce semantically equivalent code) to work around tool limitations or otherwise allow the formal verification to work. Most patches are used to mark functions as `noinline` so they will not be inlined when compiling from C to LLVM, and so we can prove them correct and re-use these proofs when functions are called. All verification is done after patching the code.

This particular patch adds in the constant verification code. This test code uses symbols that are not visible outside the `p384.c` file, so the test code needs to be inside this file. The simple solution is to patch it in, and then invoke the test function from the existing test suite.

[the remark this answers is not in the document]
apetcher-amazon: To answer your question about how the test is executed: it is patched in to the test suite by the patch on line 20 of this file, and then then the entire test suite is executed on line 30.

Coq/EC/ECEqProof.v
apetcher-amazon: Added comments. I also removed the second `Hypothesis` because it is not necessary, and also probably not true for the conversion routines supplied by fiat-crypto and s2n-bignum.

Coq/EC/ECEqProof.v
apetcher-amazon: Added a comment.

Coq/EC/Util.v
apetcher-amazon: Lemmas like this are only used to help me structure proofs of other facts. In the case of this lemma, if I have a goal like `x1 - y1 = x2 - y2`, I can apply this theorem to replace it with goals `x1 = x2` and `y1 = y2`. I could accomplish the same thing with Ltac or rewriting (in some cases), but applying a theorem is often simpler.

Coq/EC/EC_P384_Abstract_Model_equiv.v
apetcher-amazon: I added some comments and re-organized the file a bit. Many of these variables/hypotheses are internal to the proof, and they are replaced with concrete terms and proofs in ECEqProof.v. Let me know if more documentation is needed to make all this clear.

khieta: Thanks, the new re-org and comments are useful.

SAW/scripts/x86_64/entrypoint_check.sh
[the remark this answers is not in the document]
khieta: But the code patched in here is C code, right? How does that relate to the Coq validation functions?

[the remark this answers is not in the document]
apetcher-amazon: The patched in C functions are verified against a Cryptol spec using SAW. This is done in `EC_P384_validate.saw`, and the Cryptol spec was added to `EC_mul.cry` (e.g. `validate_base_table`). The Cryptol spec is converted to Gallina functions with the same name, and we prove the required correctness properties of these functions in Coq. We also assume that `validate_base_table` was executed on the pre-computed table on line 648 of `EcEqProof.v`, and this assumption is justified by the fact that we executed a C function that is proven to behave in the same way as this function.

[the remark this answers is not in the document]
khieta: Got it, thanks for the clarification!

pennyannn (approved):

PR #143, opened 2024-04-23
Add unbounded AES-GCM verification

RyanGlScott:
This proves AES-GCM functions, building on top of the `cryptol-specs` work in https://github.com/GaloisInc/cryptol-specs/pull/72 and the `saw-script` work in https://github.com/GaloisInc/saw-script/pull/2037. This supersedes previous efforts in #80 and #139.

Notes:

* This requires a newer version of SAW to work than what `aws-lc-verification`'s CI currently uses, so I needed to update the SAW binary distribution as a result. This version of SAW uses a newer version of Cryptol which overhauls the module system, and as a result, the Cryptol code in the `aws-lc-verification` repo (some of which is unrelated to the AES-GCM proofs) needed to be refactored slightly to work with the newer version of Cryptol. (See the "[Update Cryptol code to work with latest Cryptol/cryptol-specs](https://github.com/awslabs/aws-lc-verification/commit/032ffbc8469f989469210396ceb40ad08b98b0be)" commit for the details.)
* The bulk of the proofs are found in the [Add unbounded AES-GCM verification.](https://github.com/awslabs/aws-lc-verification/commit/06f825f3fdcdea6e1a272c8ff91e8c53948fdf56) commit, with some minor tweaks of my own in subsequent commits to make the proofs succeed in a more reasonable amount of time.
* Some of the AES-GCM proofs use Z3's Constrained Horn Clause (CHC) feature, which is buggy on Z3-4.8.8. `aws-lc-verification`'s CI is currently using Z3-4.8.8, so I upgrade it to a more recent version.


README.md
```diff
@@ -40,6 +40,7 @@ AWS libcrypto includes many cryptographic algorithm implementations for several
 | [SHA-2](SPEC.md#SHA-2) | 384 | EVP_DigestInit, EVP_DigestUpdate, EVP_DigestFinal | neoverse-n1, neoverse-v1 | NoEngine, MemCorrect, ArmSpecGap, ToolGap | SAW, NSym |
 | [HMAC](SPEC.md#HMAC-with-SHA-384) | with <nobr>SHA-384</nobr> | HMAC_CTX_init, HMAC_Init_ex, HMAC_Update, HMAC_Final, HMAC | SandyBridge+ | NoEngine, MemCorrect, InitZero, NoInline, CRYPTO_once_Correct | SAW |
 | [<nobr>AES-KW(P)</nobr>](SPEC.md#AES-KWP) | 256 | AES_wrap_key, AES_unwrap_key, AES_wrap_key_padded, AES_unwrap_key_padded | SandyBridge+ | InputLength, MemCorrect, NoInline | SAW |
+| [<nobr>AES-GCM</nobr>](SPEC.md#AES-GCM) | 256 | gcm_ghash_avx, aes_hw_ctr32_encrypt_blocks, aesni_gcm_encrypt, aesni_gcm_decrypt | SandyBridge+ | MemCorrect, NoInline | SAW |
```
RyanGlScott, line 43: @apetcher-amazon: The AES-GCM proofs are documented here in the `README.md`...

SPEC.md
```diff
@@ -35,6 +35,17 @@ The AES_(un)wrap_key_* functions are verified to have the following properties r
...
+| aesni_gcm_decrypt | <ul><li>The input length is a symbolic integer k such that k < 2^36.</li><li>The key parameter points to a 32-byte AES-256 key.</li><li>The hash table parameter points to values precomputed by calling `get_Htable` on the AES-256 key.</li></ul> | <ul><li>If k >= 96, then the first k bytes of the output buffer will point to the ciphertext produced by the GCM-AD_K function as defined by the AES-GCM specification. If k < 96, then the output buffer will not be modified.</li><li>If k >= 96, then the Xi buffer will point to the updated Xi value computed by the `aesni_gcm_decrypt` Cryptol function. If k < 96, then the Xi buffer will not be modified.</li><li>If k >= 96, then the initialization vector buffer will point to the updated initialization vector value computed by the `aesni_gcm_decrypt` Cryptol function. If k < 96, then the initialization vector will not be updated.</li><li>The return value will be equal to 96 * floor(k / 96).</li></ul> |
```
RyanGlScott, line 47: ...as well as here in the `SPEC.md`.

SAW/scripts/x86_64/release_jobs.sh
```diff
@@ -8,11 +8,7 @@ saw proof/SHA512/verify-SHA384-x86.saw
 saw proof/SHA512/verify-SHA512-x86.saw
 saw proof/HMAC/verify-HMAC.saw
 saw proof/KDF/verify-HKDF.saw
-# The proof for aesni_gcm_en[de]crypt requires having more than 6 function arguments.
-# Disabling the following proof due to lack of support of such a scenario in SAW.
-# The existing proof works on AWS-LC commit 7bdae64
-# (https://github.com/aws/aws-lc/commit/7bdae649ee6a723a5197e27ec3770c37f6a64e85) and before
-# cd proof/AES && go run AES-GCM-check-entrypoint.go
+(cd proof/AES && go run AES-GCM-check-entrypoint.go)
```
pennyannn, line 11: The parallel.py script is not very clever and it assumes a specific shape of command. Could you try changing the line to (i.e. removing the parentheses)
```
cd proof/AES && go run AES-GCM-check-entrypoint.go
```

RyanGlScott: Oops, my mistake. I've removed the parentheses.

RyanGlScott:
Every invocation of SAW is failing with something like this:

```
Proof results for ['/usr/bin/time', 'saw', 'proof/SHA512/verify-SHA384-x86.saw']: 0.002012973999967471s
saw: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by saw)
saw: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by saw)
saw: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by saw)
Command exited with non-zero status 1
0.00user 0.00system 0:00.00elapsed ?%CPU (0avgtext+0avgdata 720maxresident)k
0inputs+0outputs (0major+64minor)pagefaults 0swaps

Exit 1
```

I think this might be happening because the version of SAW binary located at the new `SAW_URL` is built for Ubuntu 22.04, but the Docker image is based on Ubuntu 20.04 (and therefore uses an older version of `glibc` than what the SAW binary supports). We should replace the `SAW_URL` binary with a Ubuntu 20.04 build of SAW, such as the one located at the `saw-1.1.0.99-ubuntu-20.04-X64 (GHC 9.2.8)` link [here](https://github.com/GaloisInc/saw-script/actions/runs/8785544039?pr=2037). @pennyannn, can you help me do this?

pennyannn:
@RyanGlScott , Sorry about that! I've uploaded a new SAW for Ubuntu20.04 at `https://saw-builds.s3.us-west-2.amazonaws.com/saw-1.1.0.99-2024-04-23-48dcbf7bc-PR2037-Linux-x86_64.tar.gz`.

SAW/proof/AES/AES-GCM.saw
```diff
@@ -297,7 +294,119 @@ llvm_verify m "EVP_DecryptUpdate"
...
+      w4_unint_yices ["aesni_gcm_encrypt", "aes_ctr32_encrypt_blocks_array", "aes_hw_encrypt", "gcm_ghash_blocks_array", "gcm_polyval", "pmult", "pmod", "arrayLookupUnint", "arrayUpdateUnint", "arrayCopyUnint", "arrayConstantUnint"];
+    } else if is_Xi_post then do {
+      simplify (addsimps EncryptUpdate_slice_thms empty_ss);
+      w4_unint_z3_using "qfufbv" ["aesni_gcm_encrypt", "aes_ctr32_encrypt_blocks_array", "aes_hw_encrypt", "gcm_ghash_blocks_array", "gcm_polyval", "pmult", "pmod", "arrayLookupUnint", "arrayUpdateUnint", "arrayCopyUnint", "arrayConstantUnint"];
+    } else do {
+      w4_unint_z3 ["aesni_gcm_encrypt", "aes_ctr32_encrypt_blocks_array", "aes_hw_encrypt", "gcm_ghash_blocks_array", "gcm_polyval", "pmult", "pmod", "arrayLookupUnint", "arrayUpdateUnint", "arrayCopyUnint", "arrayConstantUnint"];
+    };
+  });
+
+let do_prove = false;
```
pennyannn, line 370: `let do_prove = false` will make proofs after it assumed instead of proved, correct?

RyanGlScott: My apologies, this was left over from a debugging session. I've removed this now.

pennyannn:
@RyanGlScott for the error in the `saw-x86_64` job, that is very weird. If there is an error in SAW proofs, then `parallel.py` will print it. I suspect it might be caused by large memory usage. The script assumes a rough memory usage per job, but if AES-GCM uses way more than that, the total usage might exceed machine limit. I will try to see if I can reproduce it on my machine. 
I agree with Adam that maybe it is easiest to separate out the AES-GCM proof. We can move the files that AES-KW(P) depends on into AES-KW. Then the AES directory becomes AES-GCM.

RyanGlScott:
> I agree with Adam that maybe it is easiest to separate out the AES-GCM proof. We can move the files that AES-KW(P) depends on into AES-KW. Then the AES directory becomes AES-GCM.

That sounds reasonable to me. I'll prepare a separate PR for that, since it will likely be more self-contained than this one.

I don't want to abandon this PR just yet, as it will be useful should we ever want to migrate the non-AES-GCM proofs over to the latest SAW. I'll leave it open for now.

RyanGlScott:
I've opened #144, an alternative version of this PR which splits out the AES-GCM proofs into something more self-contained.

pennyannn:
An update on the `saw-x86_64` workflow failure. I tried to reproduce the problem and found that the culprit is the proof for `HMAC_Init_ex` in the HMAC proof. The Z3 process generated from that proof seems to be stuck for a long time and my guess is that Github Runner just kills the whole job (therefore no results are printed to the stdout/stderr). 

The high-level explanation is that this new version of SAW or Z3 does not work with the `HMAC_Init_ex` proof.

RyanGlScott:
Ah, good catch. Indeed, testing locally, it appears that the use of Z3-4.8.14 is (at least partially) to blame, as the `SHAUpdate_maintains_SHAState_eq_thm` lemma takes much, much longer to prove with 4.8.14 than it does with 4.8.8. Sigh.

pennyannn:
Correction. In my previous experiment, I used the new SAW but old Z3-4.8.8. 
When using new SAW and new Z3-4.8.14, here're the results:
```
Succeeded fast: SHA384, SHA512, AES_KW and AES_KW(P)
Succeeded: AES-GCM (about 50mins)
Stuck: HMAC, HKDF, P384, ECDH, and ECDSA
```
It will be a non-trivial amount of work to fix all these proofs.

PR #144, opened 2024-04-24
Add unbounded AES-GCM verification, isolate newer SAW from other proofs

RyanGlScott:
This proves unbounded AES-GCM functions, building on top of the cryptol-specs work in https://github.com/GaloisInc/cryptol-specs/pull/72 and the saw-script work in https://github.com/GaloisInc/saw-script/pull/2037. This supersedes previous efforts in https://github.com/awslabs/aws-lc-verification/pull/80, https://github.com/awslabs/aws-lc-verification/pull/139, and #143.

This requires a newer version of SAW to work than what aws-lc-verification's CI currently uses, so I isolated these proofs into their own SAW scripts, Cryptol specs, shell scripts, Docker image, and CI action. Moreover, some of the AES-GCM proofs use Z3's Constrained Horn Clause (CHC) feature, which is buggy on Z3-4.8.8. aws-lc-verification's other CI jobs are currently using Z3-4.8.8, so I specifically chose a newer version (Z3-4.8.14) for the new, AES-GCM–specific CI job.


RyanGlScott:
Ah, I mistakenly introduced some changes to `SAW/proof/common/helpers.saw` which require a newer version of SAW, but weren't properly isolated to only affect AES-GCM. I'll push a fix momentarily.

RyanGlScott:
At long last, the CI finally passes! Ready for review, @apetcher-amazon and @pennyannn.

apetcher-amazon:
I have a few high-level comments before I examine the details.

1. A big thing that is missing is an indication of how well the proof can be parallelized. Unless I'm missing something, the current scripts appear to only check a single pair of `mres` and `res_mres` values because `AES-GCM-check-entrypoint.go` just falls through to `AES_GCM.saw`. You should be able to modify the Go scripts to execute `AES_GCM.saw` in parallel with multiple parameter values. Can you try this so we can get some indication of how much RAM is needed and how well we can parallelize the proof? You don't need to run the SAW script on all parameter values---just show how many parameter values we can check in parallel. The runner that we are using for the CI has 64 GB of RAM. 
2. In SPEC.md, you should describe the spec of the top-level functions like `EVP_EncryptUpdate` and `EVP_EncryptFinal`. There is no need to include specs for internal functions.
3. The Cryptol spec for AES-GCM is very hard to follow. Can you organize/simplify it a bit and/or add some comments. It needs to be easy for someone to examine it and agree that it describes (some implementation-specific variant of) AES-GCM.
4. In general, the code could be improved with more comments and some slight tweaks to organization. Please look over the code and make sure that magic numbers are explained and defined using constants(e.g. why length is a 28-bit integer), variable names are clear or described using documentation (e.g. meaning of `mres` and `res_mres` are not clear from the name), etc. Also, move large proof scripts that do not need to be inspected out of top-level files like `AES-GCM.saw` so it is easier to see what is proved.

RyanGlScott:
> 1. A big thing that is missing is an indication of how well the proof can be parallelized.

Yes, a fair point. I can experiment locally to see how well we can parallelize it, although my local machine only has 32 GB of RAM.

> 2. In SPEC.md, you should describe the spec of the top-level functions like `EVP_EncryptUpdate` and `EVP_EncryptFinal`. There is no need to include specs for internal functions.

Ah, good to know. I'll do that.

> 3. The Cryptol spec for AES-GCM is very hard to follow. Can you organize/simplify it a bit and/or add some comments. It needs to be easy for someone to examine it and agree that it describes (some implementation-specific variant of) AES-GCM.

Yes, quite so. One of the challenges here is that much of the Cryptol code checked in here is not meant to reflect the official NIST AES-GCM specification so much as it is an aid for SAW to help identify the loop structure of the C code itself, so it is written in a much more "low-level" style than what you'd find in the NIST spec. The Cryptol code that is closer to the NIST spec is found in https://github.com/GaloisInc/cryptol-specs/pull/72 (in `AES_256_GCM.cry`).

That being said, the comments could definitely explain all this better. I'll do that and add some more commentary where I can.

> 4. In general, the code could be improved with more comments and some slight tweaks to organization. Please look over the code and make sure that magic numbers are explained and defined using constants(e.g. why length is a 28-bit integer), variable names are clear or described using documentation (e.g. meaning of `mres` and `res_mres` are not clear from the name), etc. Also, move large proof scripts that do not need to be inspected out of top-level files like `AES-GCM.saw` so it is easier to see what is proved.

Sure, will do.

pennyannn:
I want to confirm if my understanding is correct. In this proof, the bulk processing functions `aesni_gcm_encrypt` and `aesni_gcm_decrypt` are verified with arbitrary input length, therefore they are unbounded proofs. 

For all other functions, they are all bounded proofs serving some purpose, and this includes proofs for the top-level EVP functions as well. For example, `aes_hw_ctr32_encrypt_blocks` is only verified with number of blocks smaller than `MIN_BULK_BLOCKS = 18`. This is because this function is only used when there aren't enough message blocks to do one loop iteration of bulk processing. Likewise for GHASH functions like `gcm_ghash_avx`.

For top-level EVP function, I believe only one set of verification parameter is chosen for the proof, so it is also a bounded proof. The override matching to bulk functions `aesni_gcm_encrypt` and `aesni_gcm_decrypt` will automatically figure out what the message block length is. 

Is my understanding correct?

If so, could you add some comments to the proof command for the top-level EVP functions (the various `llvm_verify` for `EVP_EncryptUpdate` and `EVP_DecryptUpdate`) explaining the meaning of the input numbers. What do they refer to and why we are choosing this set of parameter?

Dockerfile.saw_x86_aes_gcm
```diff
@@ -0,0 +1,27 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+
+FROM ubuntu:20.04
```
pennyannn, line 5: I'm a bit confused. I thought we need ubuntu:22.04 to run the new SAW but this dockerfile uses ubuntu:20.04, and the CI still passes?

pennyannn:
Wow, this is so cool! All workflow pass and apparently the proofs now run faster even though we are checking more combinations of parameters! I wonder if it is because you removed the bounded proofs. I will start reviewing. Thank you so much for the help, @RyanGlScott !

SAW/proof/AES/GHASH.saw
```diff
@@ -77,33 +100,75 @@ gcm_gmult_avx_ov <- llvm_verify_x86 m "../../build/x86/crypto/crypto_test" "gcm_
...
+      simplify (addsimps [gcm_ghash_avx_thm] empty_ss);
+      goal_eval_unint ["pmult", "pmod", "gcm_polyval"];
+      simplify (addsimps xor_slice_append_thms basic_ss);
+      simplify (addsimps slice_slice_thms empty_ss);
+      simplify (addsimps xor_slice_append_thms basic_ss);
+      simplify (addsimps concat_assoc_0_thms empty_ss);
+      simplify (addsimps concat_assoc_1_thms empty_ss);
+      simplify (addsimps concat_assoc_2_thms empty_ss);
+      goal_eval_unint ["pmult", "pmod", "gcm_polyval"];
+      goal_eval_unint ["pmult", "pmod", "gcm_polyval"];
```
pennyannn, line 135: I wonder what's going on here that we need to do `goal_eval_unint ["pmult", "pmod", "gcm_polyval"];` so many times.

SPEC.md
```diff
@@ -35,6 +35,17 @@ The AES_(un)wrap_key_* functions are verified to have the following properties r
...
 
+## AES-GCM
+
+The EVP_* functions are verified to have the following properties related to AES-GCM using AES-256. These functions are defined in [NIST SP 800-38D](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38d.pdf). For more detailed specifications, see [AES-GCM.saw](SAW/proof/AES/AES-GCM.saw).
+
+| Function  | Preconditions |  Postconditions |
+| ---------------| -------------| --------------- |
+| EVP_CipherInit_ex | <ul><li>The global variable `OPENSSL_ia32cap_P` exists and points to a value that [disables `AVX512[F|DQ|BW|VL]` instructions](https://www.openssl.org/docs/manmaster/man3/OPENSSL_ia32cap.html).</li><li>The `ctx` parameter's `cipher` field is null.</li><li>The `cipher` parameter is a valid `EVP_CIPHER` such as the value returned by `EVP_aes_256_gcm()`.</li></ul> | <ul><li>The global variable `OPENSSL_ia32cap_P` continues to point to the same value.</li><li>The `ctx` field is initialized for AES-GCM using AES-256.</li><li>The `ctx` parameter's `cipher_data` field is initialized such that `mres` is `0`, `iv_set` is `1`, and `taglen` is `-1`.</li></ul> |
```
pennyannn, line 44: There seems to be some formatting issue with this table.

pennyannn (changes requested):
I have reviewed the workflow, shell scripts and most of the SAW files (excluding goal-rewrites files). I will send another review for the Cryptol files.

SPEC.md
```diff
@@ -35,6 +35,17 @@ The AES_(un)wrap_key_* functions are verified to have the following properties r
...
+| EVP_DecryptFinal_ex | <ul><li>The `ctx` field is initialized for AES-GCM using AES-256.</li><li>The `ctx` field's `cipher` parameter is a valid `EVP_CIPHER` such as the value returned by `EVP_aes_256_gcm()`.</li><li>The `ctx` parameter's `cipher_data` field is initialized such that `iv_set` is `1` and `taglen` is `AES_BLOCK_SIZE`.</li><li>The `ctx` parameter's `buf` field is allocated.</li></ul> | <ul><li>The `ctx` parameter's `buf` field points to the result of calling the `cipher_final ctx` Cryptol function.</li><li>The `ctx` parameter's `cipher_data` field is set such that the `mres` value is the same as before calling the function, `iv_set` is `1` if the initial value of the `ctx` parameter's `buf` field is equal to the result of calling the `cipher_final ctx` Cryptol functio (and `0` otherwise), and `taglen` is `AES_BLOCK_SIZE`.</li><li>The `out_len` parameter points to 0.</li><li>If the initial value of the `ctx` parameter's `buf` field is equal to the result of calling the `cipher_final ctx` Cryptol function, then the return value will be 1. Otherwise, the return value will be 0.</li></ul> |
```
apetcher-amazon, line 47: typo: "functio"

README.md
```diff
@@ -40,6 +42,7 @@ AWS libcrypto includes many cryptographic algorithm implementations for several
 | [SHA-2](SPEC.md#SHA-2) | 384 | EVP_DigestInit, EVP_DigestUpdate, EVP_DigestFinal | neoverse-n1, neoverse-v1 | NoEngine, MemCorrect, ArmSpecGap, ToolGap | SAW, NSym |
 | [HMAC](SPEC.md#HMAC-with-SHA-384) | with <nobr>SHA-384</nobr> | HMAC_CTX_init, HMAC_Init_ex, HMAC_Update, HMAC_Final, HMAC | SandyBridge+ | NoEngine, MemCorrect, InitZero, NoInline, CRYPTO_once_Correct | SAW |
 | [<nobr>AES-KW(P)</nobr>](SPEC.md#AES-KWP) | 256 | AES_wrap_key, AES_unwrap_key, AES_wrap_key_padded, AES_unwrap_key_padded | SandyBridge+ | InputLength, MemCorrect, NoInline | SAW |
+| [<nobr>AES-GCM</nobr>](SPEC.md#AES-GCM) | 256 | gcm_ghash_avx, aes_hw_ctr32_encrypt_blocks, aesni_gcm_encrypt, aesni_gcm_decrypt | SandyBridge+ | MemCorrect, NoInline | SAW |
```
apetcher-amazon, line 45: Change these function names to the EVP functions listed in SPEC.md.

Also, add caveats for any assumptions/limitations/etc associated with this proof. Here are the ones that I could think of:
* AES/GCM is verified using an implementation-specific Cryptol spec describing an optimized form of AES/GCM. 
* Only verified for cases where whole blocks are encrypted/decrypted---i.e. length passed to EVP_{Encrypt/Decrypt}Update is a multiple of 16. (This is true because we are verifying the case where mres=0 and res_mres=0. We verify other cases, too, but they are not as easy to characterize. We can remove this caveat if we set up the CI to cover all cases). 
* Supplying additional data (AD) is not verified. In order to verify this case, we would need to develop SAW proofs for EVP_{Encrypt/Decrypt}Update where the output buffer pointer is null. 
* Only verified for 12-byte IVs and 16-byte tags.

SAW/proof/AES/evp-function-specs.saw
```diff
@@ -42,45 +42,116 @@ let EVP_CipherInit_ex_spec enc = do {
...
   points_to_EVP_AES_GCM_CTX cipher_data_ptr ctx mres {{ 1 : [32] }} 0xffffffff;
 
   ctx_ptr <- crucible_alloc (llvm_struct "struct.evp_cipher_ctx_st");
   points_to_evp_cipher_ctx_st ctx_ptr cipher_ptr cipher_data_ptr enc;
 
-  (in_, in_ptr) <- ptr_to_fresh_readonly "in" (llvm_array len (llvm_int 8));
-  out_ptr <- crucible_alloc (llvm_array len (llvm_int 8));
+  len_28 <- mk_len;
+  let len = {{ (len_28 # (`res_mres - `mres)): [32] }};
+  crucible_precond {{ len < 2 ^^ 31 }};
```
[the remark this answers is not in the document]
apetcher-amazon: Also, if the AWS-LC implementation allows lengths larger than this, then please add a caveat to the README file stating that the implementation is only verified for input lengths less than `2^^31` bytes.

SPEC.md
RyanGlScott: Ugh, GitHub's markdown parser for tables really doesn't like the fact that I used `|` characters in the phrase "`AVX512[F|DQ|BW|VL]`". I'll change this to something else (e.g., "`AVX512{F,DQ,BW,VL}`") in the hopes that that makes it format better.

SAW/proof/AES/GHASH.saw
RyanGlScott: I don't recall why this was set up this way, but I've confirmed that you only need to call `goal_eval_unint [...]` once. I'll remove the other calls, which are redundant.

SAW/proof/AES/AES-GCM-check-entrypoint.go
```diff
@@ -34,12 +35,15 @@ func main() {
...
 	num_parallel_process := int(math.Floor((float64(total_memory) / float64(memory_used_per_test))))
 	log.Printf("System has %d bytes of memory, running %d jobs in parallel", total_memory, num_parallel_process)
-	for i := selectcheck_range_start; i <= selectcheck_range_end; i++ {
-		wg.Add(1)
-		saw_template := "verify-AES-GCM-selectcheck-template.txt"
-		placeholder_name := "TARGET_LEN_PLACEHOLDER"
-		go utility.CreateAndRunSawScript(saw_template, []string{}, placeholder_name, i, &wg)
-		utility.Wait(&process_count, num_parallel_process, &wg)
+	for i := 0; i < 2; i++ {
+		for j := 0; i < 2; i++ {
```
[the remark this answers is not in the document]
RyanGlScott: Good catch. Unfortunately, now that I've attempted a run with all four cases in parallel, I see that the SAW runners do indeed run out of memory. This suggests that we should stick to two cases in parallel. Do you have a preference for which two cases we'd want to prioritize?

SAW/proof/AES/helpers.saw
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+*/
+
+// Using experimental proof command "crucible_llvm_verify_x86"
+enable_experimental;
+
+let llvm_verify_fixpoint_x86' m elf func globals pathsat loop_func_term spec tactic = if do_prove
```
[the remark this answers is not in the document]
apetcher-amazon: Other proofs use `do_prove` to structure the proof and avoid proving the same fact multiple times. This comes up when there are dependencies between algorithms/proofs (e.g. HMAC depends on SHA-2). That isn't the case here, but I didn't consider the value of leaving it in for debugging convenience. I'm okay with leaving it as is.

SAW/proof/AES/AES-GCM-check-entrypoint.go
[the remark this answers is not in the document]
apetcher-amazon: Definitely leave in (0, 0), so we can claim that AES/GCM is verified when the input is whole blocks. If you can think of another case that might cause problems (e.g. because it uses more RAM or takes longer), then also check that one. Otherwise, any other case is fine.

SAW/proof/AES/AESNI-GCM.saw
```diff
@@ -7,47 +7,45 @@
...
-  let min_size = eval_size {| (1 + 2 * encT) * 6 * AES_BLOCK_SIZE |};
-
-  (in_, in_ptr) <- ptr_to_fresh_readonly "in" (llvm_array len (llvm_int 8));
-  out_ptr <- crucible_alloc (llvm_array len (llvm_int 8));
+// An unbounded specification for `aesni_gcm_encrypt` (if `enc == 1`) or
+// `aesni_gcm_decrypt` (if `enc == 0`). The length is symbolic, although it
+// is constrained to be less than 2 ^^ 36.
+let aesni_gcm_cipher_array_spec enc = do {
+  len <- llvm_fresh_var "len" (llvm_int 64);
+  crucible_precond {{ len < 2 ^^ 36 }};
```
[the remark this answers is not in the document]
RyanGlScott: This ultimately comes from [this line](https://github.com/GaloisInc/aws-lc/blob/98cac48f2fb9bdac41a83b8f2c9a263015cb87a6/crypto/fipsmodule/modes/asm/aesni-gcm-x86_64.pl#L109) of the x86-64 code in `aws-lc`, I believe, as this constraint is necessary to avoid integer overflow in the `add` instruction. I'll add a comment here to that effect.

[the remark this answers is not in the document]
pennyannn: That's interesting. Shouldn't this count as a memory safety issue? @apetcher-amazon, the C implementation doesn't check `len < 2 ^ 36`. So in theory, one can supply this function with a message exceeding that length and cause integer overflow. (One might argue that a message of that length might not appear in practice, though)

SAW/proof/AES/evp-function-specs.saw
```diff
@@ -42,45 +42,116 @@ let EVP_CipherInit_ex_spec enc = do {
...
+
+/*
+ * These specifications for EVP_{Encrypt,Decrypt}Final_ex are very similar, but
+ * with a handful of key differences:
+ *
+ * * EVP_EncryptFinal_ex_spec expects `buf` to be written to after calling the function, but
+ *   EVP_DecryptFinal_ex_spec does not.
+ *
+ * * EVP_EncryptFinal_ex_spec expects an initial taglen value of -1, whereas
+ *   EVP_DecryptFinal_ex_spec expects an initial taglen value of AES_BLOCK_SIZE.
```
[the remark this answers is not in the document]
RyanGlScott: Yes, good catch. I've observed that I can change the uses of `0xffffffff` to `AES_BLOCK_SIZE` in the specifications for `EVP_{Encrypt,Decrypt}Update` and `EVP_EncryptFinal` and they will still succeed. That being said, `EVP_CipherInit_ex` still appears to `taglen` to `0xffffffff`...

SAW/proof/AES/evp-function-specs.saw
```diff
@@ -42,45 +42,116 @@ let EVP_CipherInit_ex_spec enc = do {
...
   ctx_ptr <- crucible_alloc (llvm_struct "struct.evp_cipher_ctx_st");
   points_to_evp_cipher_ctx_st ctx_ptr cipher_ptr cipher_data_ptr enc;
 
-  (in_, in_ptr) <- ptr_to_fresh_readonly "in" (llvm_array len (llvm_int 8));
-  out_ptr <- crucible_alloc (llvm_array len (llvm_int 8));
+  len_28 <- mk_len;
+  let len = {{ (len_28 # (`res_mres - `mres)): [32] }};
+  crucible_precond {{ len < 2 ^^ 31 }};
+  crucible_precond {{ ctx.len < `TOTAL_MESSAGE_MAX_LENGTH }};
+  crucible_precond {{ ctx.len + (0 # len) < `TOTAL_MESSAGE_MAX_LENGTH }};
```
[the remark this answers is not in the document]
RyanGlScott: I believe this constraint ultimately arises from [this check in the code](https://github.com/GaloisInc/aws-lc/blob/98cac48f2fb9bdac41a83b8f2c9a263015cb87a6/crypto/fipsmodule/modes/gcm.c#L519-L523). (Note that `((UINT64_C(1) << 36) - 32)` is the same value as `TOTAL_MESSAGE_MAX_LENGTH`.) I'll note this in a comment.

SAW/proof/AES/AESNI-GCM.saw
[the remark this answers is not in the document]
pennyannn: Oh, I'm wrong. The C code does have such a [constraint](https://github.com/GaloisInc/aws-lc/blob/98cac48f2fb9bdac41a83b8f2c9a263015cb87a6/crypto/fipsmodule/modes/gcm.c#L519) as is shown in another comment from Ryan.

SAW/spec/AES/AES-GCM-unbounded.cry
```diff
@@ -0,0 +1,617 @@
...
+      (gcm_ghash_block H Xi (join (arrayRangeLookup blocks (i * 16))))
+  else
+    Xi
+
+
+// An unbounded specification for EVP_EncryptUpdate. This performs the update
+// phase of the GCM-AD_E algorithm described in Section 7.1 of NIST SP 800-38D.
+// Unlike the presentation in NIST SP 800-38D, this heavily uses SMT arrays,
+// which better facilitates an unbounded proof in SAW.
+aes_gcm_encrypt_update : AES_GCM_Ctx -> Array[64][8] -> [64] -> Array[64][8] -> (AES_GCM_Ctx, Array[64][8])
```
pennyannn, line 459: This definition is more closely related to the implementation than the NIST specification. How about saying something like the following:
```
// An unbounded specification for EVP_EncryptUpdate. This function performs an update to the `ctx` with input message `in` with length `len` and output the encrypted message as `out`.
```

SAW/spec/AES/AES-GCM-unbounded.cry
```diff
@@ -0,0 +1,617 @@
...
+  where
+    enc_byte = (arrayLookup in offset) ^ (eki @ j)
+
+
+// An unbounded specification for EVP_DecryptUpdate. This performs the update
+// phase of the GCM-AD_K algorithm described in Section 7.2 of NIST SP 800-38D.
+// Unlike the presentation in NIST SP 800-38D, this heavily uses SMT arrays,
+// which better facilitates an unbounded proof in SAW.
+aes_gcm_decrypt_update : AES_GCM_Ctx -> Array[64][8] -> [64] -> Array[64][8] -> (AES_GCM_Ctx, Array[64][8])
+aes_gcm_decrypt_update ctx in len out =
```
pennyannn, line 542: Similarly for this function. I think it is more of a rewrite of the C implementation for `EVP_DecryptUpdate` than is related to the specification.

SAW/spec/AES/AES-GCM-unbounded.cry
```diff
@@ -0,0 +1,617 @@
...
+  enc_block2 = prefetch4 # prefetch5
+  enc_block3 = prefetch6 # prefetch7
+  enc_block4 = prefetch8 # prefetch9
+  enc_block5 = drop Z3
+  blocks = [enc_block0, enc_block1, enc_block2, enc_block3, enc_block4, enc_block5]
+  (next_spill_Z3, next_loop_Xi, next_Z0) = gcm_polyval_pmult4_impl key spill_Z3 (drop loop_Xi) (drop Z0) blocks
+
+
+// An unbounded specification for aesni_gcm_encrypt.
+aesni_gcm_encrypt :
```
pennyannn, line 375: Point out that this definition represent the bulk encrypt assembly. The bulk encryption phase processes 6 block chunks, and this definition also does 6 blocks in each recursive call.

SAW/spec/AES/AES-GCM-unbounded.cry
```diff
@@ -0,0 +1,617 @@
...
+    then aesni_gcm_encrypt len in out key iv ctr Xi (loop_index + 1) next_out next_Xi
+    else (current_out, current_Xi, (join iv) # ((join ctr) + (drop (6 * ((len / 16) / 6)))))
+ where
+  in_blocks = split (join (arrayRangeLookup`{n=96} in (loop_index * 96)))
+  out_blocks = aes_ctr32_encrypt_blocks (join key) (join iv) ((join ctr) + (drop (6 * loop_index))) in_blocks
+  next_out = arrayRangeUpdate current_out (loop_index * 96) (split (join out_blocks))
+  next_Xi = gcm_ghash_blocks (gcm_init_H (join (get_H' key))) current_Xi out_blocks
+
+// An unbounded specification for aesni_gcm_decrypt.
+aesni_gcm_decrypt :
```
pennyannn, line 398: Similarly for this definition. Point out the bulk decryption.

SAW/spec/AES/AES-GCM-unbounded.cry
```diff
@@ -0,0 +1,617 @@
...
+  LoopTuple
+aesni_gcm_encrypt_impl_loop len in out key ivec ctr Xi loop_tuple =
+  if 6 * (get_loop_index loop_tuple) + 24 <= len / 16
+    then
+      aesni_gcm_encrypt_impl_loop len in out key ivec ctr Xi
+        (aesni_gcm_encrypt_impl_loop_step len in out key ivec ctr Xi loop_tuple)
+    else
+      loop_tuple
+
+aesni_gcm_encrypt_impl_loop_step :
```
pennyannn, line 232: This function looks like part of the inductive invariant. Is that the case? Could you create a comment for it?

SAW/spec/AES/AES-GCM-unbounded.cry
```diff
@@ -0,0 +1,617 @@
...
+  LoopTuple ->
+  LoopTuple
+aesni_gcm_encrypt_impl_loop_step len in out key ivec ctr Xi
+  ( loop_out
+  , prefetch9, prefetch8, prefetch7, prefetch6, prefetch5, prefetch4, prefetch3, prefetch2
+  , spill_Z3
+  , prefetch1, prefetch0
+  , inout5, inout4, inout3, inout2, inout1, inout0, loop_Xi, Z3, Z0, T1
+  , prefetch_offset, ctr_least_byte, loop_index
+  ) = ( arrayRangeUpdate loop_out (loop_index * 96 + 192) (split (join [out_block0, out_block1, out_block2, out_block3, out_block4, out_block5]))
```
pennyannn, line 249: I couldn't figure out where the number `192` is coming from. Sorry I kept asking about these magical numbers. I wonder if you know in case.

SAW/spec/AES/AES-GCM-unbounded.cry
```diff
@@ -0,0 +1,617 @@
...
+  Array[64][8] ->
+  Array[64][8] ->
+  [32][8] ->
+  [12][8] ->
+  [4][8] ->
+  [16][8] ->
+  LoopTuple ->
+  LoopTuple
+aesni_gcm_encrypt_impl_loop len in out key ivec ctr Xi loop_tuple =
+  if 6 * (get_loop_index loop_tuple) + 24 <= len / 16
```
pennyannn, line 225: What does the number 24 refer to?

SAW/proof/AES/goal-rewrites.saw
```diff
@@ -228,101 +227,728 @@ let gcm_polyval_thms =
...
+        current_out
+        (loop_index * 96)
+        (split (join (aes_ctr32_encrypt_blocks`{6} (join key) (join iv) ((join ctr) + (drop (6 * loop_index)))
+          (split (join (arrayRangeLookup in (loop_index * 96))))))))
+      (loop_index * 96 + 96)
+      (split (join (aes_ctr32_encrypt_blocks`{6} (join key) (join iv) ((join ctr) + (drop (6 * (loop_index + 1))))
+        (split (join (arrayRangeLookup in (loop_index * 96 + 96)))))))
+}};
+
+aesni_gcm_encrypt_impl_loop_hyp <- prove_theorem assume_unsat (rewrite (cryptol_ss ()) (unfold_term ["aesni_gcm_encrypt_impl_loop_stmt"] {{ aesni_gcm_encrypt_impl_loop_stmt }}));
```
pennyannn, line 499: This theorem is assumed and then used in the proof for `aesni_gcm_encrypt_impl_loop_thm`. I think this is probably some kind of a mocked induction proof. Is that correct? Could you add a comment to describe how it works? Also, I don't know if it should be mentioned in the caveats. @apetcher-amazon

SAW/proof/AES/goal-rewrites.saw
```diff
@@ -228,101 +227,728 @@ let gcm_polyval_thms =
...
+            , 0 # (swap8 ((join iv) # ((join ctr) + (drop (6 * loop_index)))))
+            , 96 * loop_index
+            , 0 # ((swap8 (join ctr)) + (((drop (6 * loop_index)) : [8]) # 0))
+            , loop_index
+            ))))
+    else
+      (aesni_gcm_decrypt len in out key iv ctr Xi loop_index current_out current_Xi))
+  }};
+
+aesni_gcm_decrypt_impl_loop_hyp <- prove_theorem assume_unsat aesni_gcm_decrypt_impl_loop_stmt;
```
pennyannn, line 638: Same here.

SAW/proof/AES/goal-rewrites.saw
```diff
@@ -228,101 +227,728 @@ let gcm_polyval_thms =
...
   })
-  (rewrite (cryptol_ss ()) {{ \(x : [120]) (y : [8]) key0 key1 -> aesenc ((x # y) ^ key0) key1 == aesenc (key0 ^ (x # y)) key1 }});
-aesenclast_thm <- prove_print
+  aesni_gcm_decrypt_impl_loop_stmt;
+
+let aesni_gcm_encrypt_Yi_stmt = rewrite (cryptol_ss ()) {{ \len in out key iv ctr Xi i out' Xi' ->
+  (aesni_gcm_encrypt len in out key iv ctr Xi i out' Xi').2
+  == (join iv) # ((join ctr) + (drop (6 * ((len / 16) / 6))))
+}};
+aesni_gcm_encrypt_Yi_hyp <- prove_theorem assume_unsat aesni_gcm_encrypt_Yi_stmt;
```
pennyannn, line 707: Also here.

SAW/proof/AES/goal-rewrites.saw
```diff
@@ -228,101 +227,728 @@ let gcm_polyval_thms =
...
   })
-  (rewrite (cryptol_ss ()) {{ \x y z -> aesenclast x (y ^ z) == (aesenclast x y) ^ z }});
-aesenc_aesenclast_thm <- prove_print
+  aesni_gcm_encrypt_Yi_stmt;
+
+let aesni_gcm_decrypt_Yi_stmt = rewrite (cryptol_ss ()) {{ \len in out key iv ctr Xi i out' Xi' ->
+  (aesni_gcm_decrypt len in out key iv ctr Xi i out' Xi').2
+  == (join iv) # ((join ctr) + (drop (6 * ((len / 16) / 6))))
+}};
+aesni_gcm_decrypt_Yi_hyp <- prove_theorem assume_unsat aesni_gcm_decrypt_Yi_stmt;
```
pennyannn, line 721: Also here.

pennyannn (commented):
I finished reviewing goal-rewrites and the Cryptol files.

SAW/spec/AES/AES-GCM-unbounded.cry
RyanGlScott: I believe `192` comes from [this part](https://github.com/GaloisInc/aws-lc/blob/98cac48f2fb9bdac41a83b8f2c9a263015cb87a6/crypto/fipsmodule/modes/asm/aesni-gcm-x86_64.pl#L498-L504) of the x86-64 implementation of `aesni_gcm_encrypt` (thankfully, this part has an accompanying comment). I'll mention this here in the Cryptol comments.

SAW/proof/AES/AESNI-GCM.saw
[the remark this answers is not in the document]
apetcher-amazon: Also see page 8 of the [NIST spec](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38d.pdf) that expresses a similar constraint on the length. In the spec, `len <= 2^36 - 32`, but checking for `len < 2^36` in this implementation accomplishes the same thing, since `len` is always a multiple of 144. Though it would be great if you could check my reasoning here, and add this information in the comment.

SAW/proof/AES/evp-function-specs.saw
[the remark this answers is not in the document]
apetcher-amazon: I think we are better off using `0xffffffff` for `taglen` in all specs except for `EVP_DecryptUpdate`, but we are missing the formal verification of one function. During decryption, there should be a call like `EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_GCM_SET_TAG, 16, tag)` before the call to `EVP_DecryptFinal` that sets the expected tag value and its length. This is when `0xffffffff` will be replaced with `AES_BLOCK_SIZE` in the context.

Can you also verify this function when called in this way? It should be trivial, since it just updates the context. While you are at it, we should probably also verify it when called to get the tag out of the context during encryption like `EVP_CIPHER_CTX_ctrl(ctx, EVP_CTRL_AEAD_GET_TAG, taglen, tag)`.

SAW/proof/AES/goal-rewrites.saw
apetcher-amazon: Yes, there should be a caveat associated with whatever is assumed here. If it is induction without a check to ensure that it is well-founded, then you can create a caveat similar to `SAWBreakpoint`.

SAW/spec/AES/AES-GCM-unbounded.cry
RyanGlScott: I'm afraid I'm not sure about this one.

SAW/proof/AES/goal-rewrites.saw
RyanGlScott: Yes, that is my understanding as well. I'll add a comment next to this use of `assume_unsat` (as well as similar uses elsewhere in this file) and add a caveat in `SPEC.md` to describe them.

SAW/proof/AES/AESNI-GCM.saw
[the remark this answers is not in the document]
RyanGlScott: You are correct—the proofs also go through with `len <= 2^^36 - 32` as the precondition. I will update this comment to mention why it suffices to check for `len < 2 ^^ 36` here (pointing to the NIST spec as well).

SAW/proof/AES/evp-function-specs.saw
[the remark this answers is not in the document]
RyanGlScott: I've added two proofs of `EVP_CIPHER_CTX_ctrl` per your suggestions. I opted not to add `EVP_CIPHER_CTX_ctrl` to `SPEC.md`, as I wasn't sure if `EVP_CIPHER_CTX_ctrl` constituted a top-level AES-GCM function. I can add it if you'd like, however.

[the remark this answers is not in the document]
apetcher-amazon: Thanks! These are `EVP_` functions, so I think it is worthwhile to add them to SPEC.md.

SAW/spec/AES/AES-GCM-unbounded.cry
RyanGlScott: Ah, I think you are right on the money. Indeed, we need to account for the three iterations here, and given that this code is already multipying by 6, we are already taking the number of blocks into account. As for why this code uses an extra iteration, I'm not sure. I don't think there is a particularly good reason for doing so, however. We can just as well write this code without needing the extra iteration:

```diff
@@ -228,7 +228,7 @@ aesni_gcm_encrypt_impl_loop :
   LoopTuple ->
   LoopTuple
 aesni_gcm_encrypt_impl_loop len in out key ivec ctr Xi loop_tuple =
-  if 6 * (get_loop_index loop_tuple) + 24 <= len / 16
+  if get_loop_index loop_tuple + 3 < (len / 16) / 6
     then
       aesni_gcm_encrypt_impl_loop len in out key ivec ctr Xi
         (aesni_gcm_encrypt_impl_loop_step len in out key ivec ctr Xi loop_tuple)
```

And similarly for `aesni_gcm_decrypt_impl_loop`:

```diff
@@ -311,7 +311,7 @@ aesni_gcm_decrypt_impl_loop :
   LoopTuple ->
   LoopTuple
 aesni_gcm_decrypt_impl_loop len in out key ivec ctr Xi loop_tuple =
-  if 6 * (get_loop_index loop_tuple) + 12 <= len / 16
+  if get_loop_index loop_tuple + 1 < (len / 16) / 6
     then
       aesni_gcm_decrypt_impl_loop len in out key ivec ctr Xi
         (aesni_gcm_decrypt_impl_loop_step len in out key ivec ctr Xi loop_tuple)
```

The proofs continue to work after making this change. I like this much better, as we can more readily explain where the magic `3`/`1` numbers come from. I've went ahead and pushed these changes (along with comments)—let me know what you think.

apetcher-amazon: I agree that the spec looks better now, but I don't understand how it is possible to prove these functions correct with both versions of the specs, since they seem to describe different behavior. Do you know the explanation for this?

RyanGlScott: I think they are equivalent, actually. We can show this through equational reasoning:


```
   if 6 * (get_loop_index loop_tuple) + 24          <= len / 16
== if 6 * (get_loop_index loop_tuple) + 6 * (3 + 1) <= len / 16         [24 == 6 * (3 + 1)]
== if 6 * (get_loop_index loop_tuple + 3 + 1)       <= len / 16         [distributive law]
== if get_loop_index loop_tuple + 3 + 1             <= (len / 16) / 6   [divide by 6]
== if get_loop_index loop_tuple + 3                 <  (len / 16) / 6   [definition of <=]
```

This assumes that `get_loop_index loop_tuple <= len` and `len < 2 ^^ 36`, which are both things that hold true in this spot.

apetcher-amazon: Oh. I didn't notice that you replaced `<=` with `<`. Now we are back to the original problem where I don't understand how this spec describes the behavior of the code. It looks to me like these functions in the implementation will process the input if it is exactly equal to the minimum size (96 bytes for decrypt and 288 bytes for encrypt). E.g. [here](https://github.com/GaloisInc/aws-lc/blob/98cac48f2fb9bdac41a83b8f2c9a263015cb87a6/crypto/fipsmodule/modes/asm/aesni-gcm-x86_64.pl#L680) is the test for encrypt that will only return if the number of bytes is strictly less than 288. But the spec in both cases appears to need an extra 6 blocks. 

There is a good chance that I'm reading the Cryptol spec or assembly code incorrectly. Do you believe that the spec and implementation behave the same way when called on exactly the minimum size input?

SAW/proof/AES/AESNI-GCM.saw
```diff
@@ -7,47 +7,51 @@
...
+// An unbounded specification for `aesni_gcm_encrypt` (if `enc == 1`) or
+// `aesni_gcm_decrypt` (if `enc == 0`). The length is symbolic, although it
+// is constrained to be less than 2 ^^ 36.
+let aesni_gcm_cipher_array_spec enc = do {
+  len <- llvm_fresh_var "len" (llvm_int 64);
+  // This constraint is required to avoid overflowing an addition with a large
+  // number in the x86-64 implementation of `aesni_gcm_encrypt`.
+  //
+  // According to NIST SP 800-38D (page 8), the requirement is that
+  // `len <= 2 ^^ 36 - 32`. Note that `len` is always a multiple of 144,
```
pennyannn, line 19: 144 -> 96

pennyannn (approved):
LGTM. I just had one minor comment.

SAW/spec/AES/AES-GCM-unbounded.cry
pennyannn: I don't know if this is helpful in understanding the problem. 
The Cryptol specification for the bulk encryption is this function called `aesni_gcm_encrypt`. The definition in discussion here called `aesni_gcm_encrypt_impl_loop` is not the top-level definition but some function for helping with the invariant proof. (See: [link1](https://github.com/RyanGlScott/aws-lc-verification/blob/rgs/aes-gcm-unbounded-isolated/SAW/proof/AES/AESNI-GCM.saw#L41) and [link2](https://github.com/RyanGlScott/aws-lc-verification/blob/rgs/aes-gcm-unbounded-isolated/SAW/proof/AES/AESNI-GCM.saw#L85))

RyanGlScott: Yes, I think this is the key point. `aesni_gcm_encrypt_impl_loop`'s key use is here:

https://github.com/awslabs/aws-lc-verification/blob/4397eb356e18b0f924a99d0e084bd50c2843c9db/SAW/proof/AES/goal-rewrites.saw#L453-L486

Note that we always call `aesni_gcm_encrypt_impl_loop_step` once after `aesni_gcm_encrypt_impl_loop` finishes, so this is likely a reason why the indexing in `aesni_gcm_encrypt_impl_loop` is "off-by-one" to account for it.

apetcher-amazon: Thanks. This explanation helps. Can you add some form of this explanation as comments in the SAW script near these expressions (for both encrypt and decrypt)?

RyanGlScott: I've added some comments about this.

apetcher-amazon (approved):

PR #147, opened 2024-05-08
Updating EC/ECDH/ECDSA for new P-384 code with common point doubling function.

apetcher-amazon:

Ticket: P128758919
AWS-LC PR: https://github.com/aws/aws-lc/pull/1567


pennyannn (commented):
LGTM. Some observation that is not related to this fix. 

The definition for `points_to_ec_method_st` seems to be out of sync with definition at https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/ec/p384.c#L1121. I wonder if we should update it to match them.

apetcher-amazon:
> LGTM. Some observation that is not related to this fix.
> 
> The definition for `points_to_ec_method_st` seems to be out of sync with definition at https://github.com/aws/aws-lc/blob/main/crypto/fipsmodule/ec/p384.c#L1121. I wonder if we should update it to match them.

Good catch. The differences aren't a problem in the current state of formal verification (e.g. the function pointers aren't used by verified functions, or the pointed-to functions are assumed to be correct, anyway). But this is a good opportunity to correct this difference, so I made the changes in this PR.

pennyannn (approved):

PR #152, opened 2024-06-11
Update EC proofs for ec-nistp refactoring

pennyannn:

AWS-LC PR: https://github.com/aws/aws-lc/pull/1602
Ticket: P133735854

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,7 @@
 [submodule "src"]
 	path = src
-	branch = main
-	url = https://github.com/aws/aws-lc.git
+	branch = ec_nistp_refactor-v2
+	url = https://github.com/dkostic/aws-lc.git
 [submodule "cryptol-specs"]
 	path = cryptol-specs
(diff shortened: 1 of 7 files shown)


.gitmodules
```diff
@@ -1,7 +1,7 @@
 [submodule "src"]
 	path = src
-	branch = main
-	url = https://github.com/aws/aws-lc.git
+	branch = ec_nistp_refactor-v2
```
robdockins, line 3: Will this target `main` again once the upstream PR lands?

robdockins (approved):
As best I can tell, the main content of this change is to move from using a function that was specific to P-384 to one that now applies to all the NIST curves.

It would be easier to follow if the commit message mentioned the high-level reason for this change. Otherwise, this seems unproblematic to me.

.gitmodules
pennyannn: Yes. We follow a rule for flipping back the submodule after every fixing PRs like this one.

PR #155, opened 2024-06-14
Update SHA proof for "moving AArch64/X86_64 dispatching to C" changes

pennyannn:

AWS-LC PR: https://github.com/aws/aws-lc/pull/1625
Ticket: V1417594860

The AWS-LC PR refactors existing SHA2 code to move AArch86/x86_64 dispatching code from within the assembly to within the C. In doing so, assembly function names are changed in the following way:

1. For x86_64, the `sha512_block_data_order` routine is now called `sha512_block_data_order_avx`
2. For AArch64, the `sha512_block_data_order` routine is now called `sha512_block_data_order_nohw` (for Graviton2) and `sha512_block_armv8` routine is now called `sha512_block_data_order_hw` (for Graviton3)

While working on the fix, I solved a problem in the proof of `EVP_DigestFinal`. Previously, the proof fails for SHA512 because the LLVM optimization introduces pointer comparisons, which is an undefined behaviour in SAW (see https://github.com/GaloisInc/saw-script/issues/1308). SAW produces the error:
  "Comparison of pointers from different allocations" 
  
In this fix, we use SAW feature `enable_lax_pointer_ordering` to allow pointer comparison introduced by LLVM optimization. In addition, this proof requires defining an override for `CRYPTO_bswap8` for the version that uses `__builtin_bswap64`:
```
static inline uint64_t CRYPTO_bswap8(uint64_t x) {
  return __builtin_bswap64(x);
}
```


SAW/proof/SHA512/SHA512.saw
```diff
@@ -43,15 +43,6 @@ EVP_DigestUpdate_array_ov <- llvm_verify m "EVP_DigestUpdate"
 
 // Verify the `EVP_DigestFinal` C function satisfies the
 // `EVP_DigestFinal_array_spec` unbounded specification.
-// TODO:
```
robdockins, line 46: This version of the proof still has this problem, right? You haven't added the `enable_lax_pointer_ordering` directive in this file.

SAW/proof/SHA512/verify-SHA384-aarch64-neoverse-v1.saw
```diff
@@ -7,4 +7,4 @@ import "../../../cryptol-specs/Primitive/Keyless/Hash/SHA384.cry";
 
 include "SHA384-defines.saw";
 include "aarch64-neoverse-v1.saw";
-include "SHA512.saw";
```
robdockins, line 10: Are there any remaining uses of the previous `SHA512.saw" file? If not, should it be removed?

README.md
```diff
@@ -39,7 +39,7 @@ AWS libcrypto includes many cryptographic algorithm implementations for several
 | Algorithm | Variants |  API Operations | Platform   | Caveats | Tech |
 | ----------| -------------| --------------- | -----------| ------------ | --------- |
 | [SHA-2](SPEC.md#SHA-2) | 384, 512 | EVP_DigestInit, EVP_DigestUpdate, EVP_DigestFinal | SandyBridge+ | NoEngine, MemCorrect | SAW |
-| [SHA-2](SPEC.md#SHA-2) | 384 | EVP_DigestInit, EVP_DigestUpdate, EVP_DigestFinal | neoverse-n1, neoverse-v1 | NoEngine, MemCorrect, ArmSpecGap, ToolGap | SAW, NSym |
+| [SHA-2](SPEC.md#SHA-2) | 384, 512 | EVP_DigestInit, EVP_DigestUpdate, EVP_DigestFinal | neoverse-n1, neoverse-v1 | NoEngine, NoInline, MemCorrect, ArmSpecGap, ToolGap | SAW, NSym |
```
robdockins, line 42: Does the using the "lax_pointer_ordering" option rise to the level of requiring another caveat? I'm inclined to say no, but the issue of pointer identity is fraught.

robdockins (approved):
Seems reasonable to me, modulo a couple of minor questions.

SAW/proof/SHA512/SHA512.saw
pennyannn: This version of the proof is for x86_64, interestingly it does not have the pointer comparison problem.

pennyannn: We used to have this comment because the same file is used for both x86_64 and aarch64 proofs. In this PR, I divided it into two files `SHA512.saw` (for x86_64) and `SHA512_aarch.saw` (for aarch64). That is why I removed the comment.

SAW/proof/SHA512/verify-SHA384-aarch64-neoverse-v1.saw
pennyannn: Yes. It is used in the x86 proofs.

README.md
pennyannn: That's a good question. Taking a look at the codebase, there is another place where `lax_pointer_ordering` is used in the EC proofs and it wasn't considered a caveat. I personally think the only reason that this could be considered a caveat is that it might hide other pointer comparison problems that are actually problems. Given that, I will add this as a caveat.

robdockins: Seems sensible to me.

darylmartin100 (approved):

PR #162, opened 2024-08-27
Upgrade SAW, Cryptol and cryptol-specs versions

pennyannn:

C11 upgrade in AWS-LC revealed a bug in older version of SAW. To enable proofs for C11, SAW needs to be upgraded. This PR is an attempt at fixing the failures in proofs when upgrading SAW and relevant tooling. 

Relevant issue: https://github.com/GaloisInc/saw-script/issues/2099
Previous attempt when integrating AES-GCM proofs: https://github.com/awslabs/aws-lc-verification/pull/143

### DIFF
--- a/Coq/docker_entrypoint.sh
+++ b/Coq/docker_entrypoint.sh
@@ -6,7 +6,7 @@
 # The container runs as root, but the git checkout was performed by another user.
 # Disable git security checks that ensure files are owned by the current user.
 git config --global --add safe.directory '*'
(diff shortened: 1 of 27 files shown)


pennyannn:
The shape of the term generated for HMAC_Init_ex theorem has changed in SAW. The subterm generated for the C code for slicing `key` and putting it back together into 64 bits BVs in calculation of `ipad` and `opad` look suspicious.

